PR #3, opened 2017-01-16
Monotonic Basin Hopping (first meta-algorithm)

darioizzo:


doc/sphinx/docs/algorithms/mbh.rst
```diff
@@ -0,0 +1,8 @@
+.. cpp_mbh
+
+Monotonic Basin Hopping (MBH) - Generalized
+===========================================================
+
+.. doxygenclass:: pagmo::mbh
+   :project: PaGMOreborn
```
bluescarni: This is wrong (project is ``pagmo`` now) and unnecessary anyway as we set the default project in the sphinx config file. It can go.

include/pagmo/algorithms/mbh.hpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+private:
+    // Delete all that we do not want to inherit from problem
+    // A - Common to all meta
+    template <typename T>
+    bool is() const = delete;
+    // population evolve(const population &pop) const = delete;
+    // template <typename Archive>
+    // void serialize(Archive &ar) = delete;
```
bluescarni: I suppose this commented out code can just be removed? (3 lines above I mean)

include/pagmo/problem.hpp
```diff
@@ -779,6 +809,16 @@ struct prob_inner final : prob_inner_base {
...
     }
+    template <typename U, typename std::enable_if<has_c_tolerance<U>::value, int>::type = 0>
+    static vector_double get_c_tol_impl(const U &value)
+    {
+        return value.get_c_tol();
+    }
+    template <typename U, typename std::enable_if<!has_c_tolerance<U>::value, int>::type = 0>
+    static vector_double get_c_tol_impl(const U &)
+    {
+        return vector_double{};
```
bluescarni: Wouldn't it be better here to just return a vector of zeroes of the correct size, rather than special casing the empty vector case below?

include/pagmo/problem.hpp
```diff
@@ -1013,6 +1058,14 @@ class problem
         if (m_nic > std::numeric_limits<decltype(m_nic)>::max() / 3u) {
             pagmo_throw(std::invalid_argument, "The number of inequality constraints is too large");
         }
+        m_c_tol = ptr()->get_c_tol();
```
bluescarni: Can we just get the c_tol() here without the if (see previous comment).

include/pagmo/problem.hpp
darioizzo: Indeed, but then we cannot call get_nec and get_nic as for the static, so we do not have the size here ...

doc/sphinx/docs/algorithms/mbh.rst
```diff
@@ -0,0 +1,7 @@
+.. cpp_mbh
+
+Monotonic Basin Hopping (MBH) - Generalized
+===========================================================
+
+.. doxygenclass:: pagmo::de
```
bluescarni: Should probably be ``pagmo::mbh`` here.

include/pagmo/algorithms/compass_search.hpp
```diff
@@ -68,8 +69,7 @@ namespace pagmo
  * a standard in the scientific computing community for exactly the reason observed
  * by Davidon: it is slow but sure'.
  *
- * @note This algorithm does not work for multi-objective problems, nor for
- * constrained optimization or stochastic problems.
+ * @note This algorithm does not work for multi-objective problems, nor for stochastic problems.
```
bluescarni: I would also rather have this notice to be put in the documentation of ``evolve()``, as the exception is thrown there. (I mean, it can be in both places, but the throwing behaviour of ``evolve()`` should be documented).

darioizzo: Ok, this will change a ton of stuff across as @note is used everywhere.

bluescarni: @darioizzo we can do it at a later stage. Just FYI moving forward, not necessary to change it now.

darioizzo: Yeps, I was delaying the issue of the evolve documentations as its missing everywhere. I added the docs for compass search to this PR

include/pagmo/algorithms/mbh.hpp
```diff
@@ -0,0 +1,374 @@
...
+     * @return evolved population
+     * @throws std::invalid_argument if the problem is multi-objective or stochastic
+     * @throws std::invalid_argument if the perturbation vector size does not equal the problem size
+     */
+    population evolve(population pop) const
+    {
+        // We store some useful variables
+        const auto &prob = pop.get_problem(); // This is a const reference, so using set_seed for example will not be
+                                              // allowed (pop.set_problem_seed is)
+        auto dim = prob.get_nx();             // This getter does not return a const reference but a copy
```
darioizzo: :) I guess it was there to remind me why above its const auto & and there is only auto ....

include/pagmo/algorithms/mbh.hpp
```diff
@@ -0,0 +1,374 @@
...
+    {
+        return m_verbosity;
+    }
+    /// Gets the perturbation vector
+    const vector_double &get_perturb() const
+    {
+        return m_perturb;
+    }
+    /// Sets the perturbation vector
+    void set_perturb(const vector_double &perturb)
```
darioizzo: Probably only for NaNs....

darioizzo: Actually not, same as in the ctor

include/pagmo/algorithms/mbh.hpp
```diff
@@ -0,0 +1,374 @@
...
+    bool has_set_verbosity() const = delete;
+
+#if __GNUC__ > 4
+    // NOTE: We delete the streaming operator overload called with mbh, otherwise the inner algo would stream
+    // NOTE: If a streaming operator is wanted for this class remove the line below and implement it.
+    friend std::ostream &operator<<(std::ostream &, const mbh &) = delete;
+#endif
+
+    unsigned int m_stop;
+    // The member m_perturb is mutable as to allow to construct mbh also using a perturbation defined as a scalar
```
darioizzo: I also do not like this behaviour ... but could not come with something better. Maybe your suggestion would improve ... not sure.. will think about it. Leaving as it is for the moment

include/pagmo/utils/constrained.hpp
```diff
@@ -53,8 +53,8 @@ namespace detail
 
 /// Tests equality constraints against some tolerance vector. Returns number of constraints satisfied and the L2 norm of
 /// the violation
-template <typename It>
-std::pair<vector_double::size_type, double> test_eq_constraints(It ceq_first, It ceq_last, It tol_first)
```
darioizzo: crash? no. Return the wrong answer? maybe, need to check ... will do

darioizzo: I checked and there is no crash. Simply the returned norm of the violation will be NaN and that particular constraint will be considered as violated.

bluescarni:
Merged!

PR #6, opened 2017-01-18
Review of problem structure and documentation

bluescarni:


include/pagmo/algorithm.hpp
```diff
@@ -69,7 +69,7 @@ namespace pagmo
 /**
  * This type trait will be \p true if \p T provides a method with
  * the following signature:
- * @code
+ * @code{.unparsed}
```
darioizzo: why not use @code{.cpp} ?

bluescarni: doxygen was being rather inconsistent about the automatic links it would generate in the code section (e.g., it would link to pagmo::vector_double correctly but it would fail to generate a link for, e.g., std::vector<vector_double>). Plus my experience with sphinx and breathe is that they mess up badly the syntax highlighting generate from doxygen code blocks.

So the idea is just to avoid having doxygen/sphinx try to parse anything and just give us a dumb code block.

include/pagmo/problem.hpp
```diff
@@ -994,44 +944,30 @@ class problem
...
+     * - the number of objectives of the UDP is zero,
+     * - the number of objectives, equality or inequality constraints is larger than an implementation-defined value,
+     * - the problem bounds are invalid (e.g., they contain NaNs, the dimensionality of the lower bounds is
+     *   different from the dimensionality of the upper bounds, etc. - note that infinite bounds are allowed),
+     * - the \p %gradient_sparsity() and \p %hessians_sparsity() methods of the UDP fail basic sanity checks (e.g.,
+     *   they return vectors with repeated indices, they contain indices exceeding the problem's dimensions, etc.).
+     * @throws unspecified any exception thrown by the invoked methods in the UDP or by memory errors in strings
+     * and standard containers.
+     *
+     * TODO c_tol documentation.
```
darioizzo: remove?

include/pagmo/problem.hpp
```diff
@@ -1272,88 +1255,104 @@ class problem
...
-     * @endcode
+     * The availability of the gradient sparsity is determined as follows:
+     * - if the UDP does not satisfy pagmo::has_gradient_sparsity, then this method will always return \p false;
+     * - if the UDP satisfies pagmo::has_gradient_sparsity but it does not satisfy
+     *   pagmo::override_has_gradient_sparsity, then this method will always return \p true;
+     * - if the UDP satisfies both pagmo::has_gradient_sparsity and pagmo::override_has_gradient_sparsity,
+     *   then this method will return the output of the \p %has_gradient_sparsity() method of the UDP.
      *
-     * in the user-defined problem
+     * Note that, regardless of what this method returns, the problem::gradient_sparsity() method will always return
```
darioizzo: ```**NOTE** ```

PR #8, opened 2017-01-26
first commit to hr sort nan problem fix

darioizzo:


include/pagmo/detail/custom_comparisons.hpp
```diff
@@ -0,0 +1,79 @@
...
+for more details.
+
+You should have received copies of the GNU General Public License and the
+GNU Lesser General Public License along with the PaGMO library.  If not,
+see https://www.gnu.org/licenses/. */
+
+#ifndef PAGMO_CUSTOM_COMPARISONS_HPP
+#define PAGMO_CUSTOM_COMPARISONS_HPP
+
+#include "../type_traits.hpp"
```
bluescarni: Can you switch the includes around? Usually we put system includes first.

include/pagmo/problem.hpp
```diff
@@ -1761,8 +1762,17 @@ class problem
...
 
-    // A small helper to check if a vector containes unique elements.
-    template <typename U>
+    // A small helper to check if a vector containes unique elements. The version for arithmetic types
+    // is also protected against possible nans
+    template <typename U, enable_if_is_floating_point<U> = 0>
+    static bool all_unique(std::vector<U> x)
+    {
+        std::sort(x.begin(), x.end(), [](const U &el1, const U &el2) { return detail::less_than_f(el1, el2); });
+        auto it = std::unique(x.begin(), x.end());
```
bluescarni: Here we have a similar problem with ``std::unique`` I believe :( ``std::unique`` seems to require a functor that behaves like an equivalence relation, but for floating-point we have the classic ``!(NaN == NaN)``.

include/pagmo/type_traits.hpp
```diff
@@ -193,6 +193,22 @@ using decay_t = typename std::decay<T>::type;
 template <bool B, typename T = void>
 using enable_if_t = typename std::enable_if<B, T>::type;
 
+/// SFINAE enabler for arithmetic types
```
bluescarni: I'd move these enablers in the detail namespace, as they are not used in any public interface that I see and they are rather simple anyway.

include/pagmo/utils/discrepancy.hpp
```diff
@@ -23,7 +23,7 @@ or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
 for more details.
 
 You should have received copies of the GNU General Public License and the
-GNU Lesser General Public License along with the PaGMO library.  If not,
+GNU Lesser General Public Li`cense along with the PaGMO library.  If not,
```
bluescarni: Looks like a typo.

include/pagmo/utils/discrepancy.hpp
```diff
@@ -81,7 +82,8 @@ inline std::vector<double> sample_from_simplex(std::vector<double> in)
         pagmo_throw(std::invalid_argument, "Input vector must have all elements in [0,1]");
     }
     if (in.size() > 0u) {
-        std::sort(in.begin(), in.end());
+        std::sort(in.begin(), in.end(),
+                  [](const double &el1, const double &el2) { return detail::less_than_f(el1, el2); });
```
bluescarni: Doesn't it work here if, instead of a lambda, you use ``detail::less_than_f<double>`` directly?

include/pagmo/utils/generic.hpp
```diff
@@ -265,7 +266,7 @@ std::vector<std::vector<vector_double::size_type>> kNN(const std::vector<vector_
         std::vector<vector_double::size_type> idxs(N);
         std::iota(idxs.begin(), idxs.end(), vector_double::size_type(0u));
         std::sort(idxs.begin(), idxs.end(), [&distances](vector_double::size_type idx1, vector_double::size_type idx2) {
-            return distances[idx1] < distances[idx2];
+            return detail::less_than_f(distances[idx1], distances[idx2]);
```
bluescarni: These are integral indices, aren't they? There should be no need to use the ``less_than_f``.

include/pagmo/utils/multi_objective.hpp
```diff
@@ -390,10 +391,11 @@ std::vector<vector_double::size_type> sort_population_mo(const std::vector<vecto
     // Sort the indexes
     std::sort(retval.begin(), retval.end(),
               [&tuple, &crowding](vector_double::size_type idx1, vector_double::size_type idx2) {
-                  if (std::get<3>(tuple)[idx1] == std::get<3>(tuple)[idx2]) {     // same non domination rank
-                      return crowding[idx1] > crowding[idx2];                     // crowding distance decides
-                  } else {                                                        // different non domination ranks
-                      return std::get<3>(tuple)[idx1] < std::get<3>(tuple)[idx2]; // non domination rank decides
+                  if (std::get<3>(tuple)[idx1] == std::get<3>(tuple)[idx2]) { // same non domination rank
+
```
bluescarni: An extra line seems to have slipped in here.

include/pagmo/problem.hpp
```diff
@@ -1761,8 +1762,17 @@ class problem
         return m_ptr.get();
     }
 
-    // A small helper to check if a vector containes unique elements.
-    template <typename U>
+    // A small helper to check if a vector containes unique elements. The version for arithmetic types
```
bluescarni: ``arithmetic`` -> ``floating point``

include/pagmo/type_traits.hpp
bluescarni: ``arithmetic`` -> ``floating point``

include/pagmo/utils/discrepancy.hpp
darioizzo: indeed

include/pagmo/utils/generic.hpp
darioizzo: They are indexes, but they are ordered w.r.t. the associated  distance which is a float

include/pagmo/problem.hpp
darioizzo: I am not sure that is a requirement for unique (http://en.cppreference.com/w/cpp/algorithm/unique) seems that it just needs something that return a bool

darioizzo: So maybe the question is, do we need that nan == nan is true? Philosophy ... so for the time being I would leave as is

include/pagmo/utils/generic.hpp
bluescarni: Ah yeah you are right of course.

include/pagmo/detail/custom_comparisons.hpp
```diff
@@ -74,6 +74,23 @@ inline bool greater_than_f(T a, T b)
...
+            return a == b; // a == b
+        else
+            return false; // a == nan
+    } else {
+        if (!std::isnan(b))
+            return false; // nan == b
+        else
+            return true; // nan == nan
+    }
+}
```
bluescarni: Can this be compressed a bit? Something like:

```c++
if (!std::isnan(a) && !std::isnan(b)) {
  return a == b;
}
return std::isnan(a) && std::isnan(b);
```

bluescarni: or even
```c++
return (!std::isnan(a) && !std::isnan(b)) ? (a == b) : (std::isnan(a) && std::isnan(b));
```
if you are a ternary op fetishist :)

darioizzo: yes, i kept the same structure as the others for "readibility" bu will change it

PR #9, opened 2017-01-27
WIP Island sandbox

bluescarni:
Initial early experiments for the island class.


bluescarni:
@darioizzo this is ready to go in as far as I am concerned.

pygmo/__init__.py
```diff
@@ -113,7 +111,143 @@ def _population_init(self, prob=None, size=0, seed=None):
...
+        udi: a user-defined island (either Python or C++ - note that *udi* will be deep-copied
+          and stored inside the :class:`~pygmo.core.island` instance)
+        algo: a user-defined algorithm (either Python or C++), or an instance of :class:`~pygmo.core.algorithm`
+        pop (:class:`~pygmo.core.population`): a population
+        prob: a user-defined problem (either Python or C++), or an instance of :class:`~pygmo.core.problem`
+        size (``int``): the number of individuals
+        seed (``int``): the random seed (if not specified, it will be randomly-generated)
+
+    Raises:
+        KeyError: if the set of keyword arguments is invalid
```
bluescarni: @darioizzo the link is present in the island class description, the documentation here ends up attached at the bottom of it. If you look at the sphinx output, there will be a link just above. I think the same will be true when you do ``island?`` from ipython.

bluescarni: @darioizzo I opened #53 which contains some ideas about improvements to the pygmo documentation. I think the end result in sphinx at the moment looks nice, but I noticed there are some small inconsistencies in the ipython ``?`` syntax (e.g., documentation added from C++ can sometimes look a bit different from the docs added from Python). Perhaps we can continue the discussion in #53.

PR #12, opened 2017-02-02
Particle Swarm Optimization

darioizzo:
Several changes were necessary for the sign correctness of the algorithm and more. Resulting code is somehow not that nice ... hopefully it will do


include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+                 *  [Poli et al., 2007] http://dx.doi.org/10.1007/s11721-007-0002-0
+                 *  [Mendes et al., 2004] http://dx.doi.org/10.1109/TEVC.2004.826074
+                 *-------------------------------------------------------------------------------------*/
+                else if (m_variant == 6u) {
+                    for (decltype(dim) d = 0u; d < dim; ++d) {
+                        sum_forces = 0.;
+                        for (decltype(neighb[p].size()) n = 0u; n < neighb[p].size(); ++n) {
+                            sum_forces += drng(m_e) * acceleration_coefficient * (lbX[neighb[p][n]][d] - X[p][d]);
+                        }
+                        m_V[p][d] = m_omega * (m_V[p][d] + sum_forces / static_cast<double>(neighb[p].size()));
```
bluescarni: This ``static_cast`` is superfluous, as you are using it as a divisor with a ``double`` dividend: the compiler will take care of do the casting for you.

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+                if (gen % m_verbosity == 1u || m_verbosity == 1u) {
+                    // We compute the number of function evaluations made
+                    auto feval_count = prob.get_fevals() - fevals0;
+                    // We compute the average across the swarm of the best fitness encountered
+                    vector_double local_fits(swarm_size);
+                    for (decltype(swarm_size) i = 0u; i < swarm_size; ++i) {
+                        local_fits[i] = lbfit[i][0];
+                    }
+                    auto lb_avg = std::accumulate(local_fits.begin(), local_fits.end(), 0.)
+                                  / static_cast<double>(local_fits.size());
```
bluescarni: Same here, no need for the static cast.

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+                    auto best = local_fits[static_cast<unsigned int>(idx_best)];
+                    // We compute a measure for the average particle velocity across the swarm
+                    auto mean_velocity = 0.;
+                    for (decltype(m_V.size()) i = 0u; i < m_V.size(); ++i) {
+                        for (decltype(m_V[i].size()) j = 0u; j < m_V[i].size(); ++j) {
+                            if (ub[j] > lb[j]) {
+                                mean_velocity += std::abs(m_V[i][j] / (ub[j] - lb[j]));
+                            } // else 0
+                        }
+                        mean_velocity = mean_velocity / static_cast<double>(m_V[i].size());
```
bluescarni: Maybe can rewrite this as ``mean_velocity /= m_V[i].size()``?

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+                    vector_double local_fits(swarm_size);
+                    for (decltype(swarm_size) i = 0u; i < swarm_size; ++i) {
+                        local_fits[i] = lbfit[i][0];
+                    }
+                    auto lb_avg = std::accumulate(local_fits.begin(), local_fits.end(), 0.)
+                                  / static_cast<double>(local_fits.size());
+                    // We compute the best fitness encounterd so far across generations and across the swarm
+                    auto idx_best = std::distance(std::begin(local_fits),
+                                                  std::min_element(std::begin(local_fits), std::end(local_fits)));
+                    auto best = local_fits[static_cast<unsigned int>(idx_best)];
```
bluescarni: To be really pedantic, there's 2 things potentially problematic:

* ``std::distance`` will return a signed quantity (that's the reason to do the ``static_cast`` I presume), and this means that it could overflow the size type of ``local_fits``;
* the cast to ``unsigned`` should probably be a cast to the size type of ``local_fits``.

Unfortunately, I don't have much to offer in terms of constructive suggestions, so this code can go in without modifications as far as I am concerned. Properly fixing this would require some ugly checks, maybe you could add a comment as a FIXME or TODO for the future?

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+     *  less than or equal to sqrt(n) that evenly divides n and c = n / r'' \n
+     *  [Mohais et al., 2005] http://dx.doi.org/10.1007/11589990_80 \n
+     *  (there's an error in the description above: "smallest integer" should
+     *  instead be "highest integer")
+     *
+     *  @param[out] neighb definition of the swarm's topology
+     */
+    void initialize_topology__von(std::vector<std::vector<vector_double::size_type>> &neighb) const
+    {
+        int swarm_size = static_cast<int>(neighb.size());
```
bluescarni: Is there a reason why in this topology we are using signed ints rather than the size types of ``neighb`` etc.?

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+        int cols, rows; // lattice structure
+        int p_x, p_y;   // particle's coordinates in the lattice
+        int n_x, n_y;   // particle neighbor's coordinates in the lattice
+
+        rows = static_cast<int>(std::sqrt(swarm_size));
```
bluescarni: This looks a bit risky. Could we at least add a check on the ``swarm_size`` so we are sure its sqrt fits within the guaranteed minimum range of the integer type?

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+        while (swarm_size % rows != 0u)
```
bluescarni: Couldn't this be just written as ``rows -= swarm_size % rows`` or am I missing something?

include/pagmo/algorithms/pso.hpp
```diff
@@ -0,0 +1,875 @@
...
+        cols = swarm_size / rows;
+
+        for (decltype(swarm_size) pidx = 0u; pidx < swarm_size; pidx++) {
+            p_x = pidx % cols;
+            p_y = pidx / cols;
+
+            for (unsigned int nidx = 0u; nidx < 4u; nidx++) {
+                n_x = (p_x + vonNeumann_neighb_diff[nidx][0]) % cols;
+                if (n_x < 0)
+                    n_x = cols + n_x; // sign of remainder(%) in a division when at least one of the operands is
```
bluescarni: This comment is not true anymore as of C++11, the sign in a modulo operation is the sign of the dividend.

http://stackoverflow.com/questions/13100711/operator-modulo-change-in-c-11

pygmo/docstrings.cpp
```diff
@@ -1017,7 +1014,7 @@ The adaptation of the mutation variant is added to :class:`~pygmo.core.sade`
     variant_adptv (``int``): F and CR parameter adaptation scheme to be used (one of 1..2)
```
bluescarni: Maybe ``F`` and ``CR`` should be bracketed by asterisks, as done a couple of lines below?

include/pagmo/algorithms/pso.hpp
darioizzo: I did not manage to change it as I did in the others without having to rewrite the for loops, that is not straight forward

include/pagmo/algorithms/pso.hpp
darioizzo: boh .... do we need to answer? This code is taken from what Luis wrote and tested extensively 8 years ago. I ported many things but I am uncomfortable to change these small logics .....

include/pagmo/algorithms/pso.hpp
darioizzo: left it as is. Could not find a solution that is "decent".

PR #14, opened 2017-02-03
Hypervolume

bluescarni:


CoolRunning:
What is this?

bluescarni:
@CoolRunning is this the real life?

pygmo/core.cpp
```diff
@@ -91,6 +91,13 @@ see https://www.gnu.org/licenses/. */
 #include <pagmo/problems/translate.hpp>
 #include <pagmo/problems/zdt.hpp>
 #include <pagmo/serialization.hpp>
+#include <pagmo/utils/hv_algos/hv_algorithm.hpp>
```
bluescarni: ``#include <boost/shared_ptr.hpp>``

pygmo/core.cpp
```diff
@@ -885,8 +892,80 @@ BOOST_PYTHON_MODULE(core)
 
     moead_.def("get_seed", &moead::get_seed);
 
+    // Exposition of various structured utilities
+    // Hypervolume class
+    bp::class_<hypervolume>("hypervolume", "Hypervolume class.", bp::no_init)
+        .def(bp::init<population, const bool>((bp::arg("pop"), bp::arg("verify") = true)))
```
bluescarni: Any reason why this constructor is not in the line above, in place of the ``bp::no_init``? Also the ``const bool`` looks a bit unnecessary, ``bool`` alone is enough (in the original C++ ctor as well).

pygmo/core.cpp
```diff
@@ -885,8 +892,80 @@ BOOST_PYTHON_MODULE(core)
...
+        .def("__init__",
+             bp::make_constructor(+[](const bp::object &points,
+                                      const bool verify) { return ::new hypervolume(pygmo::to_vvd(points), verify); },
+                                  bp::default_call_policies(), (bp::arg("points"), bp::arg("verify") = true)))
+        .def("compute", +[](hypervolume &hv, const bp::object &r_point) { return hv.compute(pygmo::to_vd(r_point)); },
```
bluescarni: Here and in the other exposed methods, is it necessary that ``hypervolume &hv`` is taken by non-const reference? I am not sure if it matters in practice, but I'd prefer the reference to be ``const`` where possible.

pygmo/core.cpp
```diff
@@ -885,8 +892,80 @@ BOOST_PYTHON_MODULE(core)
...
+        .def("get_copy_points", &hypervolume::get_copy_points)
+        .def("get_points", +[](hypervolume &hv) { return pygmo::vv_to_a(hv.get_points()); })
+        .def("set_verify", &hypervolume::set_verify)
+        .def("get_verify", &hypervolume::get_verify);
+    // Hypervolume algorithms
+    bp::class_<hv_algorithm, boost::noncopyable>("_hv_algorithm", bp::no_init).def("get_name", &hv_algorithm::get_name);
+    bp::class_<hvwfg, bp::bases<hv_algorithm>>("wfg", "WFG algorithm.")
+        .def(bp::init<const unsigned int>((bp::arg("stop_dimension") = 2)));
+    bp::class_<bf_approx, bp::bases<hv_algorithm>>("bf_approx", "Bringmann-Friedrich approximated algorithm.")
+        .def(bp::init<const bool, const unsigned int, const double, const double, const double, const double,
```
bluescarni: The use of these ``const`` is a bit baffling (I imagine they were in the original C++ code as well?).

tests/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 find_package(Boost 1.55.0 REQUIRED)
 
+# We move the test files into the testing directory
+file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/hypervolume_test_data/" DESTINATION "${CMAKE_BINARY_DIR}/tests/hypervolume_test_data/")
```
bluescarni: ``CMAKE_BINARY_DIR`` should probably be ``CMAKE_CURRENT_BINARY_DIR ``.

pygmo/core.cpp
```diff
@@ -885,8 +892,80 @@ BOOST_PYTHON_MODULE(core)
 
     moead_.def("get_seed", &moead::get_seed);
 
+    // Exposition of various structured utilities
```
bluescarni: This and the other exposed classes/functions/methods need to be documented in the same way we document the rest of pygmo, with the sphinx napoleon syntax etc.

It would be good to have some tests on the python side as well (to test the exposition of the methods, not the hv functionality per-se).

doc/doxygen/Doxyfile
```diff
@@ -851,7 +851,7 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = ../../include/pagmo/external ../../include/pagmo/detail
+EXCLUDE                = ../../include/pagmo/external ../../include/pagmo/detail/cec2013_data.hpp
```
bluescarni: Do we want to exclude only ``cec2013_data.hpp`` here from doxygen, and not the whole ``detail/`` subdir? Am I reading this correctly?

pygmo/core.cpp
darioizzo: Yes, I am fighting to put the code to a decent standard, but there are so many of these things that fighting all is impossible.

pygmo/core.cpp
darioizzo: I can put const, but its only hiding the fact that the object is very changeable as its main data (the points) are declared mutable and may change if the flag copy_points is set to false

tests/CMakeLists.txt
darioizzo: but then the tests are not finding the test files, marcus expects them to be in:
```
/ root directory of the hypervolume data
    std::string input_data_dir("./hypervolume_test_data/");
```

doc/doxygen/Doxyfile
darioizzo: nah, thisis only a temp sol to explore doxygen warnings when the whole hv headers were there too.

pygmo/core.cpp
darioizzo: added

tests/CMakeLists.txt
bluescarni: Are these test files large? Could we just use the same trick as in cec2013?

pygmo/core.cpp
bluescarni: My point was that if, e.g., ``hv.compute()`` is a const method, I'd like for consistency the exposition wrapper to take ``const hypervolume &hv`` as first argument rather than ``hypervolume &hv``.

pygmo/core.cpp
bluescarni: No worries, this is not high priority.

tests/CMakeLists.txt
darioizzo: I do not like the trick very much, for cec2013 is ok as it gives the user some advantage, but these are for testing. I think for this purpose it better to have files than to add big static vectors that end up being parsed by clang, doxygen and more ....

pygmo/core.cpp
darioizzo: done

include/pagmo/utils/hv_algos/hv_hvwfg.hpp
```diff
@@ -0,0 +1,413 @@
...
+    std::vector<double> contributions(std::vector<vector_double> &points, const vector_double &r_point) const
+    {
+        std::vector<double> c;
+        c.reserve(points.size());
+
+        // Allocate the same members as for 'compute' method
+        allocate_wfg_members(points, r_point);
+
+        // Prepare the memory for first front
+        double **fr = new double *[m_max_points];
```
bluescarni: I suppose we don't care about the lack of exception safety here?

include/pagmo/utils/hv_algos/hv_hvwfg.hpp
```diff
@@ -0,0 +1,413 @@
...
+            // function for std::sort
+            // We need that in order for the cmp_points to have acces to the m_current_slice member variable.
+            std::sort(points, points + n_points, [this](double *a, double *b) { return this->cmp_points(a, b); });
+        }
+
+        double H = 0.0;
+        --m_current_slice;
+
+        if (rec_level >= m_n_frames) {
+            double **fr = new double *[m_max_points];
```
bluescarni: And here as well?

include/pagmo/utils/hv_algos/hv_hvwfg.hpp
```diff
@@ -0,0 +1,413 @@
...
+        return false;
+    }
+
+    /// Allocate the memory for the 'compute' method
+    void allocate_wfg_members(std::vector<vector_double> &points, const vector_double &r_point) const
+    {
+        m_max_points = points.size();
+        m_max_dim = r_point.size();
+
+        m_refpoint = new double[m_max_dim];
```
bluescarni: Ditto.

pygmo/core.cpp
```diff
@@ -861,8 +869,87 @@ BOOST_PYTHON_MODULE(core)
...
+    // Hypervolume class
+    bp::class_<hypervolume>("hypervolume", "Hypervolume Class")
+        .def("__init__", bp::make_constructor(+[](const population &pop) { return ::new hypervolume(pop, true); },
+                                              bp::default_call_policies(), (bp::arg("pop"))),
+             pygmo::hv_init2_docstring().c_str())
+        .def("__init__", bp::make_constructor(
+                             +[](const bp::object &points) { return ::new hypervolume(pygmo::to_vvd(points), true); },
```
bluescarni: Calling ``new`` like this can lead to memory leaks, if for instance ``to_vvd()`` throws (which it might, if it is passed a wrong object from Python). ``to_vvd()`` must be called in its own line, and the resulting object can then be passed to the ``new`` expression constructing the hypervolume.

pygmo/docstrings.cpp
```diff
@@ -1602,7 +1602,7 @@ The adaptation of the mutation variant is added to :class:`~pygmo.core.sade`
 
 Args:
     gen (``int``): number of generations
-    allowed_variants (``NumPy array or list of floats``): allowed mutation variants, each one being a number in [1, 18]
+    allowed_variants (``NumPy array or list of int``): allowed mutation variants, each one being a number in [1, 18]
```
bluescarni: I'd put only ``list`` and ``int`` in the backticks, as the convention elsewhere is that the backticks are used for Python entities (``float``, ``int``, ``True``, ``None``, etc.). Also, what kind of objects are exactly allowed in input here? Would tuples work? Other types of iterable objects?

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
 }
 
+std::string hv_init1_docstring()
+{
+    return R"(__init__(pop)
+
+Constructor from population
+
+Args:
+    pop (``pygmo.population``): a pygmo.population
```
bluescarni: I'd put a link to the pygmo docs for population here in the type specification (without the backticks as sphinx already takes care of formatting during link generation). Description might be "the input population" instead?

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+}
+
+std::string hv_init2_docstring()
+{
+    return R"(__init__(points)
+
+Constructor from points
+
+Args:
+    points (``list`` of ``arrays``): the points
```
bluescarni: Same questions as above, regarding formatting and the exact allowed types.

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+std::string hv_compute_docstring()
+{
+    return R"(hypervolume.compute(ref_point, hv_algo = auto)
+
+Computes the hypervolume with the supplied algorithm. If no algorithm
+is supplied,  then an exact hypervolume algorithm is automatically selected
+specific for the point dimension.
+
+Args:
+    ref_point (``list`` of ``arrays``): the points
```
bluescarni: Ditto.

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+    hv_algo (``pygmo.hv_algorithm``): hypervolume algorithm to be used
```
bluescarni: Doc link here?

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+std::string hv_contributions_docstring()
+{
+    return R"(hypervolume.contributions(ref_point, hv_algo = auto)
+
+This method returns the exclusive contribution to the hypervolume of every point.
+According to *hv_algo* this computation can be implemented optimally (as opposed to calling
+for :func:`~pygmo.hypervolume.exclusive` in a loop).
+
+Args:
+    ref_point (``list`` of ``arrays``): the points
```
bluescarni: Ditto, ditto (not adding any more comments about this).

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,155 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+    idx (``int``): index of the point
+    ref_point (``list`` of ``arrays``): the reference points
+    hv_algo (``pygmo.hv_algorithm``): hypervolume algorithm to be used
+
+
+Returns:
+    ``Numpy array``: the contribution of all points to the hypervolume
+
+Raises:
+    ValueError: if *ref_point* is not suitable
```
bluescarni: I'd prefer a list of cases in which ``ValueError`` is thrown, instead of multiple ``Raises`` clauses.

tests/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-find_package(Boost 1.55.0 REQUIRED)
+# We move the test files into the testing directory
+file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/hypervolume_test_data/" DESTINATION "${CMAKE_BINARY_DIR}/tests/hypervolume_test_data/")
```
bluescarni: I'd still insist on having ``CMAKE_CURRENT_BINARY_DIR`` here. Can you just print the paths of ``CMAKE_BINARY_DIR``, ``CMAKE_CURRENT_BINARY_DIR`` and see if you can alter the copy command accordingly?

The reason I insist is that it's easier to embed pagmo in other projects if we use relative paths for files etc., and the rest of the build system tries to do that.

include/pagmo/utils/hv_algos/hv_hvwfg.hpp
darioizzo: You mean memory allocation failures?
I would as a waiver for this PR as Its full of these things and I am not sure it would make sense to go through the whole code and fix these things. Also, I would have to ask you how to fix them properly ....

pygmo/docstrings.cpp
darioizzo: I guess any iterable looking at the  ```pygmo::to_vu``` code.
Looking at the docs for fitness, there we say (array-like object) is that ok to refer to any python iterable object?

pygmo/docstrings.cpp
darioizzo: :) at the moment we did not do docs for the hv_algorithms

tests/CMakeLists.txt
darioizzo: The thing is if I do it on arch then:

-- /home/dario/Documents/pagmo2/build
-- /home/dario/Documents/pagmo2/build

they are the same!! But when I did change it appveyor failed complaining the files where not found.

pygmo/docstrings.cpp
darioizzo: done

tests/CMakeLists.txt
bluescarni: That's rather weird. Could you try to do a debug commit on appveyor with the printing code in cmake to see what it outputs?

doc/sphinx/docs/cpp/utils/hypervolume.rst
```diff
@@ -0,0 +1,23 @@
...
+    :alt: hypervolume
+    :align: right
+
+The computation of hypervolumes plays an important role in solving multi-objective optimization
+problems. In pagmo we include a number of different methods and algorithms that allow
+hypervolumes to be computed efficiently at all dimensions. More information on the details
+of the algorithms implemented and their performance can be found in the publication:
+
+:Title: "Empirical performance of the approximation of the least hypervolume contributor."
+:Authors: Nowak, Krzysztof, Marcus Märtens, and Dario Izzo.
```
bluescarni: Isn't the first author formatting a bit funky here?

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,154 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+{
+    return R"(hypervolume.compute(ref_point, hv_algo = auto)
+
+Computes the hypervolume with the supplied algorithm. If no algorithm
+is supplied,  then an exact hypervolume algorithm is automatically selected
+specific for the point dimension.
+
+Args:
+    ref_point (2d-array like object): the points
+    hv_algo (``pygmo.hv_algorithm``): hypervolume algorithm to be used
```
bluescarni: Could you put here (and elsewhere, similarly) a hyperlink to ``pygmo.hv_algorithm`` instead?

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,154 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+This method returns the exclusive contribution to the hypervolume of every point.
+According to *hv_algo* this computation can be implemented optimally (as opposed to calling
+for :func:`~pygmo.hypervolume.exclusive` in a loop).
+
+Args:
+    ref_point (2d-array like object): the points
+    hv_algo (``pygmo.hv_algorithm``): hypervolume algorithm to be used
+
+Returns:
+    ``Numpy array``: the contribution of all points to the hypervolume
```
bluescarni: "1D NumPy array of float"? (here and elsewhere)

pygmo/docstrings.cpp
```diff
@@ -1938,4 +1938,154 @@ Runs the fast non dominated sorting algorithm on the input *points*
...
+
+std::string hv_exclusive_docstring()
+{
+    return R"(hypervolume.exclusive(idx, ref_point, hv_algo = auto)
+
+Computes the exclusive contribution to the hypervolume of a particular point.
+
+Args:
+    idx (``int``): index of the point
+    ref_point (array like object): the reference point
```
bluescarni: should be "array-like" for consistency with the rest.

tests/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
...
```
bluescarni: Any chance of sorting this out on appveyor with a debug commit?

bluescarni: Any chance of sorting this out on appveyor with a debug commit?

bluescarni:
@darioizzo I left some comments of minor importance. The only thing still sticking out for me is the ``CMAKE_BINARY_DIR`` stuff if you can figure out what is going on in appveyor.

Otherwise it's good for merging as far as I am concerned.

doc/sphinx/docs/python/tutorials/hypervolume.rst
```diff
@@ -0,0 +1,180 @@
...
+    >>> import pygmo as pg
+    >>> # Instantiates a 4-objectives problem
+    >>> prob = pg.problem(pg.dtlz(id=4, dim = 12, fdim=4))
+    >>> pop = pg.population(prob, 84)
+    >>> # Construct the hypervolume object
+    >>> # and get the reference point off-setted by 10 in each objective
+    >>> hv = pg.hypervolume(pop)
+    >>> offset = 5
+    >>> ref_point = [max(pop.get_f(), key = lambda x: x[it])[it] + offset for it in [0,1,2,3]]
+    >>> hv.compute(ref_point) # doctest: +SKIP
```
bluescarni: @darioizzo Just FYI, the doctest syntax support ellipsis, so if you want you can require that the first n digits of the result are correct (which is useful if you are dealing with floating-point). See:

https://github.com/bluescarni/piranha/blob/master/pyranha/math.py#L77

darioizzo: :) thanks, yes I used the ellipsis syntax before, but in this case the problem is that the whole result may be completely different as the ref_point will be kind of random

PR #16, opened 2017-02-07
Proposal for a Champion implementation

darioizzo:


include/pagmo/population.hpp
```diff
@@ -79,7 +79,7 @@ class population
      * Constructs an empty population with a pagmo::null_problem.
      * The random seed is initialised to zero.
      */
-    population() : m_prob(null_problem{}), m_e(0u), m_seed(0u)
+    population() : m_prob(null_problem{}), m_ID(), m_x(), m_f(), m_champion_x(), m_champion_f(), m_e(0u), m_seed(0u)
```
bluescarni: These member inits are redundant, aren't they?

include/pagmo/population.hpp
```diff
@@ -99,7 +99,7 @@ class population
      */
     template <typename T, generic_ctor_enabler<T> = 0>
     explicit population(T &&x, size_type pop_size = 0u, unsigned int seed = pagmo::random_device::next())
-        : m_prob(std::forward<T>(x)), m_e(seed), m_seed(seed)
+        : m_prob(std::forward<T>(x)), m_ID(), m_x(), m_f(), m_champion_x(), m_champion_f(), m_e(seed), m_seed(seed)
```
bluescarni: Ditto.

include/pagmo/population.hpp
```diff
@@ -411,6 +432,23 @@ class population
     }
 
 private:
+    // Short routine tp update the champion. Does nothing if the problem is MO
+    inline void update_champion(const vector_double& x, const vector_double& f)
```
bluescarni: The ``inline`` keyword is redundant.

include/pagmo/population.hpp
```diff
@@ -411,6 +432,23 @@ class population
     }
 
 private:
+    // Short routine tp update the champion. Does nothing if the problem is MO
```
bluescarni: ``tp`` -> ``to``

include/pagmo/population.hpp
```diff
@@ -287,6 +288,24 @@ class population
         return worst_idx(tol_vector);
     }
 
+    /// Gets the champion decision vector
+    vector_double get_champion_x() const {
+        if (m_prob.get_nobj() > 1u) {
+            pagmo_throw(std::invalid_argument,
+                        "The Champion of a population can only be extracted in single objective problems");
+        }
+        return m_champion_x;
```
bluescarni: Philosophical questions: what are the pros/cons of throwing if the champion has size zero, rather than returning an empty vector? Do we want to be silent if the user erroneously asks for a champion from an empty pop? Should we have a method that checks whether the champion exists (``has_champion()``)?

include/pagmo/population.hpp
```diff
@@ -170,6 +170,7 @@ class population
         m_x.reserve(m_x.size() + 1u);
         m_f.reserve(m_f.size() + 1u);
         // The rest is noexcept.
+        update_champion(x, f);
```
bluescarni: This function is not ``noexcept`` (see further comments below). It should be made noexcept and then moved above the comment.

include/pagmo/population.hpp
```diff
@@ -411,6 +432,23 @@ class population
...
+    inline void update_champion(const vector_double& x, const vector_double& f)
+    {
+        assert(f.size() > 0u);
+        // If the problem has multiple objectives do nothing
+        if (m_prob.get_nobj() == 1u) {
+            // If the champion does not exist create it
+            if (m_champion_x.size() == 0u) {
+                m_champion_x = x;
+                m_champion_f = f;
+            }
```
bluescarni: Wouldn't here be better to use an ``else if`` instead, so we avoid re-comparing the fitness in case this is the first champion being set?

include/pagmo/population.hpp
```diff
@@ -411,6 +432,23 @@ class population
     }
 
 private:
+    // Short routine tp update the champion. Does nothing if the problem is MO
+    inline void update_champion(const vector_double& x, const vector_double& f)
+    {
```
bluescarni: This function is not exception safe, as the copy assignment of ``f`` could throw after ``x`` has been set. A possible solution is to take the input params ``x`` and ``f`` by copy instead of const ref, and then move-assign them to ``m_champion_x`` and ``m_champion_f``.

include/pagmo/population.hpp
```diff
@@ -79,7 +79,7 @@ class population
      * Constructs an empty population with a pagmo::null_problem.
```
bluescarni: I am presuming the champion machinery will be documented in the class' description eventually?

darioizzo: This is just a draft to clarify the structure and help in making the decision on wether we want it. If we agree I would produce a fully "functional" PR

include/pagmo/population.hpp
darioizzo: They are, but I thought to put them there for consistency.

include/pagmo/population.hpp
darioizzo: I do not think it matters much. EIther solutions will do, maybe I prefer the returning empty vector as it offers more coding "flexibility"

has_champion() seems like an overkill ....

include/pagmo/population.hpp
darioizzo: does it not mean .that the function can be just copied where called rather than invoked?

bluescarni: All methods which are defined within the body of a class are automatically ``inline``.

include/pagmo/population.hpp
darioizzo: ok, but it looks like cheating, what happens if when invoking update_champion there is no memory to copy the arguments?

include/pagmo/population.hpp
darioizzo: you mean below "// The rest is noexcept" ?

include/pagmo/population.hpp
darioizzo: I added some description there now

include/pagmo/population.hpp
bluescarni: I'd rather have them gone.

doc/sphinx/docs/python/tutorials/using_population.rst
```diff
@@ -115,3 +115,28 @@ some individual decision vector:
           triggering the fitness function evaluation, but it is also possible to inject
           spurious information into the population (i.e. breaking the relation between
           decision vectors and fitness vectors imposed by the problem)
+
+The best individual in a population can be extracted as:
+
+.. doctest::
+
+    >>> # The decision vector
+    >>> pop1.get_x()[pop1.best_idx()]
```
bluescarni: I am wondering, if this is a common enough pattern, shouldn't we have a ``get_best_x()`` and ``get_best_f()`` methods that just call this line?

pygmo/core.cpp
```diff
@@ -245,6 +245,17 @@ static inline void population_prob_init(bp::class_<population> &pop_class)
         .def(bp::init<const Prob &, population::size_type, unsigned>());
 }
 
+// champion_x()
+static inline bp::object pop_champion_x_wrapper(population &pop)
+{
+    return pygmo::v_to_a(pop.champion_x());
+}
+// champion_f()
+static inline bp::object pop_champion_f_wrapper(population &pop)
```
bluescarni: Instead of writing these wrappers, could you just put them in the ``.add_property()`` call below with the usual ``+[]`` lambda trick?

tests/CMakeLists.txt
```diff
@@ -10,8 +10,6 @@ function(ADD_PAGMO_TESTCASE arg1)
     add_test(${arg1} ${arg1})
 endfunction()
 
-file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/cec2013_data/" DESTINATION "${CMAKE_BINARY_DIR}/tests/cec2013_data/")
```
bluescarni: I am a bit confused as why this stuff is here. It's because you branched before we merged cec2013 into master? Just hoping this does not come back from the graveyard :)

include/pagmo/population.hpp
bluescarni: The ``update_champion()`` function could throw when it copies its arguments upon invocation, so technically it could throw. So maybe you could modify the comment to say something like: "Update champion either throws before modfying anything, or completes successfully. The rest is noexcept."

doc/sphinx/docs/python/tutorials/using_population.rst
darioizzo: agreed, but lets leave it for a syntactic sugars PR in the future ... I guess we may change our minds once we start using pygmo

bluescarni: ok.

tests/CMakeLists.txt
darioizzo: indeed

pygmo/core.cpp
darioizzo: done

PR #19, opened 2017-02-08
WIP Pygmo iteration

bluescarni:
Another round of iterative improvements on the pygmo side.


include/pagmo/algorithms/cmaes.hpp
```diff
@@ -387,7 +387,8 @@ class cmaes
             // 6 - Update evolution paths
             ps = (1. - cs) * ps + std::sqrt(cs * (2. - cs) * mueff) * invsqrtC * (mean - meanold) / sigma;
             double hsig = 0.;
-            hsig = (ps.squaredNorm() / N / (1. - std::pow((1. - cs), (2. * counteval / static_cast<double>(lam)))))
+            hsig = (ps.squaredNorm() / N
+                    / (1. - std::pow((1. - cs), (2. * static_cast<double>(counteval) / static_cast<double>(lam)))))
```
darioizzo: a me qui non si lamentava, pensavo fosse imnplictamente fatto dall operatore *

pygmo/docstrings.cpp
```diff
@@ -629,6 +659,83 @@ of a :class:`~pygmo.core.problem`.
...
+std::string problem_c_tol_docstring()
+{
+    return R"(Constraints tolerance.
+
+This property contains an array of ``float`` that are used when checking for constraint feasibility.
+The dimension of the array is :math:`n_{ec} + n_{ic}`, and the array is zero-filled on problem
+construction.
+
+Returns:
+    1D NumPy float array: the constraints tolerance
```
darioizzo: We have to agree on some format for these numpy types. Elsewhere I used confusing types. I guess we take this problem review as baseline and change the rest?

array-like -> when its an arg taking any iterable
xD NumPy float array, xD NumPy int array in other cases

in all cases for these types we do not use the quotes?

bluescarni: I use the quotes when referring to actual Python types, such as ``int``, ``list``, ``tuple``, ``float``, and so on. Strictly speaking, in a NumPy array the type of the stored values is a runtime property of the array (i.e., both an array of ``int`` and ``float`` are instances of ``ndarray``), so that's why I am using a small sentence to describe the return type and not the backquotes.

include/pagmo/algorithms/cmaes.hpp
bluescarni: it's done by the operator, but the implicitly-generated conversion triggers a warning on GCC in debug mode.

PR #23, opened 2017-02-14
Pygmo iteration

bluescarni:
Another iteration of cleanups and new small bits.


bluescarni:
@darioizzo I think this is in a good place for merging, before I start breaking other stuff.

include/pagmo/problems/dtlz.hpp
```diff
@@ -325,7 +325,7 @@ class dtlz
         for (decltype(f.size()) i = 0u; i < f.size() - 1; ++i) {
             y += (f[i] / (1.0 + g)) * (1.0 + std::sin(3 * detail::pi() * f[i]));
         }
-        return m_fdim - y;
+        return static_cast<double>(m_fdim) - y;
```
darioizzo: I still wonder why I do not get those warnings. I have tried debug and release in clang and gcc in my architectures ... mah

pygmo/core.cpp
```diff
@@ -487,29 +425,39 @@ BOOST_PYTHON_MODULE(core)
...
+             pygmo::population_get_x_docstring().c_str())
+        .def("get_ID", +[](const population &pop) { return pygmo::v_to_a(pop.get_ID()); },
+             pygmo::population_get_ID_docstring().c_str())
         .def("get_seed", &population::get_seed, pygmo::population_get_seed_docstring().c_str());
 
     // Problem class.
     pygmo::problem_ptr = make_unique<bp::class_<problem>>("problem", pygmo::problem_docstring().c_str(), bp::no_init);
     auto &problem_class = *pygmo::problem_ptr;
-    problem_class.def(bp::init<const bp::object &>((bp::arg("prob"))))
+    problem_class.def(bp::init<const bp::object &>((bp::arg("udp"))))
```
darioizzo: udp, ok noted, will try to remember to not insert inconsistencies

pygmo/core.cpp
```diff
@@ -959,8 +923,25 @@ BOOST_PYTHON_MODULE(core)
 
     // Exposition of stand alone functions
     // Multi-objective utilities
-    bp::def("fast_non_dominated_sorting", fast_non_dominated_sorting_wrapper,
-            pygmo::fast_non_dominated_sorting_docstring().c_str(), bp::arg("points"));
+    bp::def("fast_non_dominated_sorting",
+            +[](const bp::object &x) -> bp::object {
```
darioizzo: was this not a good example of where a wrapper may be better? Long lambdas are weird no?

bluescarni: I think it's still ok, as long as it is being used only in 1 place. My general feeling is that wrappers are better when you need them in more than 1 place.

PR #27, opened 2017-02-20
Work on various improvements connected to multi-objective optimization

darioizzo:
Including:
 * moving decompose_fitness out of the meta problem
 * expanding moead with the choice of decomposition weights so that it has the same capabilities of pade
 * tutorial on multiobjective optimization


include/pagmo/utils/multi_objective.hpp
```diff
@@ -678,5 +678,109 @@ std::vector<vector_double> decomposition_weights(vector_double::size_type n_f, v
...
+    * was indicated as a decomposition method. Its value is ignored if "weighted" was indicated.
+    * @param method decomposition method: one of "weighted", "tchebycheff" or "bi"
+
+    *
+    * @return the decomposed objective.
+    *
+    * @throws std::invalid_argument if \p f, \p weight and \p ref_point have different sizes
+    * @throws std::invalid_argument if \p method is not one of "weighted", "tchebycheff" or "bi"
+    */
+vector_double decompose_objectives(const vector_double &f, const vector_double &weight, const vector_double &ref_point,
```
bluescarni: needs to be ``inline``.

include/pagmo/utils/multi_objective.hpp
```diff
@@ -678,5 +678,109 @@ std::vector<vector_double> decomposition_weights(vector_double::size_type n_f, v
     return retval;
 }
 
+/// Decomposes a vector of objectives.
+/**
+    * A vector of objectives is reduced to one only objective using a decomposition
```
bluescarni: The indentation here looks funky.

pygmo/docstrings.cpp
```diff
@@ -1625,7 +1629,8 @@ Multi Objective Evolutionary Algorithms by Decomposition (the DE variant)
 
 Raises:
     OverflowError: if *gen*, *neighbours*, *seed* or *limit* are negative or greater than an implementation-defined value
-    ValueError: if *weight_generation* is not one of 'random', 'low discrepancy', 'grid'
+    ValueError: if *decomposition* is not one of 'tchebycheff', 'weighted' or 'bi'
```
bluescarni: Could you replace this with a list?

darioizzo: Changed. But I think there are cases and this was one where the list is as easily readable as enumerating them

PR #30, opened 2017-02-22
Nsga 2

darioizzo:


include/pagmo/population.hpp
```diff
@@ -167,11 +167,29 @@ class population
...
+     * @param x decision vector to be added to the population.
+     * @param f fitness vector corresponding to the decision vector
+     *
+     * @throws unspecified any exception thrown by memory errors in standard containers.
+     */
+    void push_back(const vector_double &x, const vector_double &f)
+    {
         // Prepare quantities to be appended to the internal vectors.
         const auto new_id = std::uniform_int_distribution<unsigned long long>()(m_e);
         auto x_copy(x);
```
bluescarni: You need to copy ``f`` and move ``f_copy``. What happens right now in the line ``push_back(std::move(f))`` is that ``f`` is being cast to ``const &&vector_double``, so there's not move happening (it's a const rvalue reference), and a copy is made in the ``push_back()`` method, which could throw.

include/pagmo/algorithms/nsga2.hpp
```diff
@@ -0,0 +1,571 @@
...
+{
+public:
+    /// Single entry of the log (gen, fevals, ideal_point)
+    typedef std::tuple<unsigned int, unsigned long long, vector_double> log_line_type;
+    /// The log
+    typedef std::vector<log_line_type> log_type;
+
+    /// Constructor
+    /**
+    * Constructs the NSGA II user defined algorithm
```
bluescarni: Can you put a ``.`` at the end?

include/pagmo/algorithms/nsga2.hpp
```diff
@@ -0,0 +1,571 @@
...
+        // particular algorithm.
+        if (prob.is_stochastic()) {
+            pagmo_throw(std::invalid_argument,
+                        "The problem appears to be stochastic " + get_name() + " cannot deal with it");
+        }
+        if (prob.get_nc() != 0u) {
+            pagmo_throw(std::invalid_argument, "Non linear constraints detected in " + prob.get_name() + " instance. "
+                                                   + get_name() + " cannot deal with them.");
+        }
+        if (prob.get_nf() < 2u) {
```
bluescarni: This error message seems a bit needlessly complicated. Can't it be something like "this is a multiobjective algortihm and the problem prob.get_name() is single-objective"?

include/pagmo/algorithms/nsga2.hpp
```diff
@@ -0,0 +1,571 @@
...
+     * @throws unspecified any exception thrown by the serialization of the UDP and of primitive types.
+     */
+    template <typename Archive>
+    void serialize(Archive &ar)
+    {
+        ar(m_gen, m_cr, m_eta_c, m_m, m_eta_m, m_e, m_int_dim, m_seed, m_verbosity, m_log);
+    }
+
+private:
+    vector_double::size_type tournament_selection(vector_double::size_type idx1, vector_double::size_type idx2,
```
bluescarni: If these helper methods do not require access to data in ``this``, you could just make them ``static``.

pygmo/core.cpp
```diff
@@ -834,6 +835,26 @@ BOOST_PYTHON_MODULE(core)
...
+               +[](const nsga2 &a) -> bp::list {
+                   bp::list retval;
+                   for (const auto &t : a.get_log()) {
+                       retval.append(bp::make_tuple(std::get<0>(t), std::get<1>(t), pygmo::v_to_a(std::get<2>(t))));
+                   }
+                   return retval;
+               },
+               pygmo::nsga2_get_log_docstring().c_str());
+
+    nsga2_.def("get_seed", &nsga2::get_seed);
```
bluescarni: Could you put a docstring for this one as well?

pygmo/docstrings.cpp
```diff
@@ -1608,6 +1608,73 @@ See also the docs of the relevant C++ method :cpp:func:`pagmo::sade::get_log`.
 )";
 }
 
+std::string nsga2_docstring()
+{
+    return R"(__init__(gen = 1, cr = 0.95, eta_c = 10, m = 0.01, eta_m = 10, int_dim = 0, seed = random)
+
+Non dominated Sorting Genetic Algorithm (NSGA-II)
```
bluescarni: ``.``

include/pagmo/population.hpp
darioizzo: over my head ... the use of std::move is still obscure, maybe one day I will want you to explain :)

include/pagmo/algorithms/nsga2.hpp
darioizzo: they need the random engine m_e :(

include/pagmo/population.hpp
bluescarni: ``std::move(x)`` is just a fancy way of writing ``static_cast<T &&>(x)``. If your input type is ``const T``, the result of a move will be ``const T &&``. Move constructors need to get just a ``T &&`` as input, otherwise they are not selected by overload resolution, and the copy constructor is selected instead.

darioizzo: got it. I guess I did not know that const is carried when moving / casting

include/pagmo/algorithms/nsga2.hpp
darioizzo: done

PR #37, opened 2017-03-08
Unconstrain

darioizzo:
This meta-problem will contain what in PaGMO Legacy was available through two distinct classes death_penalty and con2unc.

In time more methods could be added and the class be extended.


include/pagmo/problems/unconstrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+            && method != "ignore_o") {
+            pagmo_throw(std::invalid_argument, "The method " + method + " is not supported (did you mispell?)");
+        }
+        // 4 - We throw if a non empty weight vector is passed but the method weghted is not selected
+        if (weights.size() != 0u && method != "weighted") {
+            pagmo_throw(std::invalid_argument,
+                        "The weight vector needs to be empty to use the unconstrain method " + method);
+        }
+        // 5 - We store the method in a more efficient enum type and the number of objectives of the orginal udp
+        std::map<std::string, method_type> my_map = {{"death penalty", method_type::DEATH},
```
bluescarni: Since this map is used in at least one other place, maybe it would be better to have it stored as a static const variable, perhaps with the usual template trick.

include/pagmo/problems/unconstrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+    // (unconstrain::hessians_sparsity) and referred to a problem with one objective
+    bool has_hessians_sparsity() const = delete;
+    std::vector<sparsity_pattern> hessians_sparsity() const = delete;
+    // No need for these the default implementation of these in pagmo::problem already returns zero.
+    vector_double::size_type get_nec() const = delete;
+    vector_double::size_type get_nic() const = delete;
+
+    /// types of unconstrain methods
+    enum class method_type {
+        DEATH = 0,    ///< "death penalty"
```
bluescarni: You can avoid providing explicit integral initializers in the enum: if they are not specified, the compiler will do exactly what you are doing here.

pygmo/core.cpp
```diff
@@ -646,7 +651,8 @@ BOOST_PYTHON_MODULE(core)
     pygmo::expose_problem<tu_test_problem>("_tu_test_problem", "A thread unsafe test problem.");
     // Null problem.
     auto np = pygmo::expose_problem<null_problem>("null_problem", pygmo::null_problem_docstring().c_str());
-    np.def(bp::init<vector_double::size_type>((bp::arg("nobj"))));
+    np.def(bp::init<vector_double::size_type, vector_double::size_type, vector_double::size_type>(
```
bluescarni: Is this gonna work if one specifies only ``nec`` in the constructor on the Python side? I feel like ``nobj`` should have a default value as well, but then I don't know if it clashes with the def ctor already exposed by ``expose_problem``.

include/pagmo/problems/unconstrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+    using ctor_enabler
+        = enable_if_t<std::is_constructible<problem, T &&>::value && !std::is_same<uncvref_t<T>, problem>::value, int>;
+
+public:
+    /// Default constructor
+    /**
+     * The default constructor will initialize a pagmo::null_problem unconstrained via the death penalty method.
+     */
+    unconstrain()
+        : problem(null_problem{2, 3, 4}), m_method(method_type::DEATH), m_weights(), m_nobj(2), m_nec(3), m_nic(4),
```
darioizzo: ah thats new in c++11 right?

include/pagmo/problems/unconstrain.hpp
darioizzo: and in any case implicit conversion between the enums value and the int is not allowed anymore right?

include/pagmo/problems/unconstrain.hpp
bluescarni: yep, pretty handy.

pygmo/core.cpp
darioizzo: That was my worry. But I tried now and seems to work .... I put nobj to 1 by default. No idea which default constructor will be called

pygmo/docstrings.cpp
```diff
@@ -1087,7 +1087,7 @@ The best known solution for the )"
 
 std::string translate_docstring()
 {
-    return R"(__init__(udp = null_problem(), translation = [0.])
+    return R"(__init__(udp, translation)
```
darioizzo: Because we do not allow default values. Those are only chosen if the default constructor is used. I guess we intend for the user to not assume any default values and assign explicitly them. The default constructor should not be called.

include/pagmo/problems/unconstrain.hpp
darioizzo: its only called once, the other is just the opposite

pygmo/docstrings.cpp
bluescarni: We do allow default values, it's also written in the documentation below, and it mirrors what happens in C++ as well. I don't see a reason to hide this fact from the user.

darioizzo: pg.problem(pg.translate(translation = [1.])) will not work, thats what I meant

bluescarni: Yes, that's why I added the blurb at the bottom about permitting certain sets of arguments for the constructor (2 or 3 + 1 optional). It seemed the easier way to document the actual behaviour of the class without going around re-implementing constructors on the Python side (unfortunately I don't have any better idea that does not involve considerable effort to emulate the behaviour of C++ functions in Python with respect to overloading).

At least with the def argument in the docstring we are documenting which problem is initialised in case the def ctor is called.

bluescarni: And which def translation vector as well.

darioizzo: k

pygmo/docstrings.cpp
```diff
@@ -2526,6 +2526,61 @@ used for Tchebycheff and boundary interception decomposition methods.
...
+
+* no arguments,
+* two mandatory arguments and one optional arguments.
+
+Any other combination of arguments will raise an exception.
+
+Args:
+    udp: a user-defined problem (either C++ or Python - note that *udp* will be deep-copied
+      and stored inside the :class:`~pygmo.core.unconstrained` instance)
+    method (``str``): a string containing the unconstrain method chosen, one of [``'death penalty'``, ``'kuri'``, ``'weighted'``, ``'ignore_c'``, ``'ignore_o'``]
```
darioizzo: I think the visualization is clearer like this as it separates with commas keywords

bluescarni:
@darioizzo looks good to me 👍

PR #42, opened 2017-03-14
The cec2009 problem suite (multi-objective and constrained)

darioizzo:
The last of the CEC problem suite. Should now give pagmo2 the possibility to be tested with all sorts of problems. We maybe still lack a good suite with gradients and hessians, but hock_shittkowsky_71 may do for the time being.


doc/sphinx/docs/cpp/problems/cec2006.rst
```diff
@@ -1,5 +1,5 @@
-CEC 2006 Problem Suite (box-bound, unconstrained)
-====================================================
+CEC 2006 Problem Suite (single-objective, constrained)
```
bluescarni: Is the C++ documentation for cec2009 missing?

include/pagmo/problems/cec2009.hpp
```diff
@@ -0,0 +1,873 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "../detail/constants.hpp"
+#include "../exceptions.hpp"
+#include "../problem.hpp" // needed for cereal registration macro
+#include "../types.hpp"
+
+#define MYSIGN(x) ((x) > 0 ? 1.0 : -1.0)
```
bluescarni: Could we make this just a private static function inside the class? Or in the detail namespace at least (in class would be better if possible)?

include/pagmo/problems/cec2009.hpp
```diff
@@ -0,0 +1,873 @@
...
+                ub[0] = 1.0;
+                lb[1] = 0.0;
+                ub[1] = 1.0;
+                for (decltype(m_dim) i = 2u; i < m_dim; ++i) {
+                    lb[i] = -2.0;
+                    ub[i] = 2.0;
+                }
+            }
+        }
+        return std::pair<vector_double, vector_double>(std::move(lb), std::move(ub));
```
bluescarni: If you want this can just be ``return std::make_pair(std::move(lb), std::move(ub));``.

doc/sphinx/docs/cpp/problems/cec2006.rst
darioizzo: as usual forgotten file:)

include/pagmo/problems/cec2009.hpp
darioizzo: sure. why the hatred for macros? namespace pollution?

bluescarni: * if we include some 3rd party headers which have ``MYSIGN`` defined, we gonna run into issues
* the name of the macro is too short and too common, the risk of name collision is high
* it's just a function, there's not a single reason why this should not be an inline function (if it needed to access information available only to macros, e.g., line number or file name, it would be ok, but it doesn't)

bluescarni: BTW there's also a nice ``std::signbit`` function that does a sensible thing in presence of NaN:

http://en.cppreference.com/w/cpp/numeric/math/signbit

(yes I know this is 3rd party code, just for general info :) )

include/pagmo/problems/cec2009.hpp
```diff
@@ -0,0 +1,877 @@
...
+       &cec2009::UF6, &cec2009::UF7, &cec2009::UF8, &cec2009::UF9, &cec2009::UF10};
+
+template <typename T>
+const std::vector<typename cec2009_statics<T>::func_ptr> cec2009_statics<T>::m_c_ptr
+    = {&cec2009::CF1, &cec2009::CF2, &cec2009::CF3, &cec2009::CF4, &cec2009::CF5,
+       &cec2009::CF6, &cec2009::CF7, &cec2009::CF8, &cec2009::CF9, &cec2009::CF10};
+} // namespace detail
+} // namespace pagmo
+
+#undef sgn
```
bluescarni: No need for this any more.

darioizzo: damn it!

bluescarni:
👍 I'll wait for CI then merge.

Issue #40, opened 2017-03-10
Split pygmo compilation


bluescarni:
As we keep on adding functionality, pygmo compilation becomes more expensive both in terms of runtime and memory usage. We should split up ``core.cpp`` in separate parts in order to curb memory usage and allow to recompile only the parts that actually need to be recompiled.

Note that we have to take care of additional NumPy initialization complexity when we split the module:

http://docs.scipy.org/doc/numpy/reference/c-api.array.html#importing-the-api

darioizzo:
Is this going to be a problem only for the CI?
At the moment all my machines do not run into troubles.

bluescarni:
It's taking about 4GB of memory for a debug build of pygmo, which I believe is getting close to travis' limit. There's also the fact that every change apart from the docstrings triggers a full rebuild, making it harder to debug as it takes quite a while to compile.

Now that pagmo is header only, changes in the core c++ bits are anyway going to trigger a full rebuild in many cases, but at least the exposition bits I believe could benefit from some splitting up.

bluescarni:
Fixed in #55.

PR #55, opened 2017-03-26
CI iteration

bluescarni:


bluescarni:
@darioizzo I'd have this merged before moving onto Windows. It includes the split of core.cpp (see #40), various fixes for missing ``inline`` uncovered by the split, and Pygmo testing on Linux/OSX.

PR #59, opened 2017-03-29
Morecov

darioizzo:
Paying the iron price

### DIFF
--- a/include/pagmo/detail/custom_comparisons.hpp
+++ b/include/pagmo/detail/custom_comparisons.hpp
@@ -122,7 +122,6 @@ struct hash_vf {
         return retval;
     }
 };
-
 } // end of detail namespace
 } // end of pagmo namespace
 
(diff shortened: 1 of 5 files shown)


darioizzo:
Good to go for me

include/pagmo/detail/custom_comparisons.hpp
```diff
@@ -90,7 +90,7 @@ inline bool equal_to_f(T a, T b)
 }
 
 // equal_to_vf than compares vectors of floating point types considering nan==nan
-template <typename T>
+template <typename T, detail::enable_if_is_floating_point<T> = 0>
```
bluescarni: The ``enable_if`` stuff is typically used in template specialisations, if you really want to have a check here it should go as a static assert in the body of the class:

```static_assert(std::is_floating_point<T>::value, "This class can be used only with floating-point types.");```

darioizzo: is it not going to work like that? or its just a matter of style?

bluescarni: It's not gonna do anything different from the static assertion, and it is more opaque. If this was a template specialisation, you could use it to pick another specialisation after you have disabled the invalid one, but in this case there's only one implementation.

PR #60, opened 2017-03-29
Pygmo windows

bluescarni:


pygmo/core.cpp
```diff
@@ -419,48 +419,48 @@ BOOST_PYTHON_MODULE(core)
         .def("__deepcopy__", &pygmo::generic_deepcopy_wrapper<population>)
         .def_pickle(population_pickle_suite())
         .def("push_back",
-             +[](population &pop, const bp::object &x, const bp::object &f) {
+             pygmo::lcast([](population &pop, const bp::object &x, const bp::object &f) {
```
darioizzo: should we also in core.cpp be ```using namespace pygmo```?

PR #61, opened 2017-04-02
Problem iteration

bluescarni:
Some work on the problem class from initial experience on nlopt integration. I'll try to describe the 3 commits here.

The first bit is easy: remove ``safe_cast`` in favour of ``boost::numeric_cast``, since we now depend on Boost anyway. Code removed, life is good.

The second bit implements some additional checking on the return values of ``gradient_sparsity()`` and ``hessians_sparsity()``. Specifically, we now check also that the returned value's dimension(s) is consistent with the dimension(s) recorded as problem members during problem construction. This was prompted by the realisation that, with an UDP which is badly behaved enough, we can trigger out-of-bounds memory accesses when mapping a sparse gradient to a dense one. This is because, for instance ``gradient()`` checks that the returned value is consistent with ``m_gs_dim`` but ``gradient_sparsity()`` previously didn't (so potentially the problem previously could return a gradient sparsity pattern inconsistent with the size of the sparse gradient vector). Same for hessians.

The third bit enforces that all sparsity patterns need to be sorted lexicographically. This is what we were always doing anyway (no tests had to be altered), and in my opinion it makes it easier to reason about and handle the sparsity information. It is also consistent with many standard formats for storing sparse data.

### DIFF
--- a/doc/sphinx/docs/cpp/miscellanea/generic.rst
+++ b/doc/sphinx/docs/cpp/miscellanea/generic.rst
@@ -19,10 +19,6 @@ A number of utilities to compute quantities that are of general relevance.
 
 --------------------------------------------------------------------------
 
-.. doxygenfunction:: pagmo::safe_cast
-
---------------------------------------------------------------------------
-
 .. doxygenfunction:: pagmo::binomial_coefficient
 
 --------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


bluescarni:
My hypothesis is that the additional branching has turned some green into yellow, but I cannot really understand the codecov output.

PR #67, opened 2017-04-04
NLopt

bluescarni:


pygmo/docstrings.cpp
```diff
@@ -3601,4 +3596,416 @@ inserted via :func:`~pygmo.core.archipelago.push_back()`).
...
+            22        17.6821             11        1.78909 i
+            23        7.71103              5       0.130386 i
+            24        6.24758              4     0.00736759 i
+            25        6.23325              1    5.12547e-05 i
+            26         6.2325              0              0
+            27        6.23246              0              0
+            28        6.23246              0              0
+            29        6.23246              0              0
+            30        6.23246              0              0
+    <BLANKLINE>
```
darioizzo: no need for <BLANKLINE> since you skip the test I guess .. on the other hand is clearer like this and consistent ... probably leave it ..

PR #81, opened 2017-04-17
A first attempt at compiling python modules as CMake MODULEs, instead…

bluescarni:
… of SHARED.


bluescarni:
This is needed for supporting the manylinux builds (#82).

PR #86, opened 2017-04-19
More tutorials and web look

darioizzo:
This contains the new website look and some improvements on tutorials.


pygmo/__init__.py
```diff
@@ -231,7 +231,7 @@ def _mbh_init(self, algo=None, stop=5, perturb=1e-2, seed=None):
         # Otherwise, we attempt to create an algorithm from it. This will
         # work if algo is an exposed C++ algorithm or a Python UDA.
         algo_arg = algorithm(algo)
-    if type(perturb) is float:
+    if type(perturb) is float or type(perturb) is int:
```
bluescarni: This is a bit tricky. There's two issues here:

* passing a ``numpy.float64`` (which is essentially a C ``double``) will fail this test, as ``numpy.float64 != float``. This can be amended by using ``isinstance(perturb,float)``, which works for both ``float64`` and ``float`` (for some reason, ``float64`` derives from ``float`` apparently). I don't think this is just an academic issue, as the ``perturb`` parameter might conceivably come out of some calculation involving numpy from users' code (numpy turns all floats into ``float64`` when you place them in an array, if I recall correctly).
* Python 2 has 2 integer types, ``int`` and ``long``, so if you pass a ``long`` the test will fail on Python 2. A solution that works in both 2 and 3 is here: https://github.com/esa/pagmo2/blob/master/pygmo/__init__.py#L308

I don't know what's the best way of proceeding here. What we want to test here, I presume, is whether ``perturb`` is a scalar or not. There's a function in numpy which might be useful https://docs.scipy.org/doc/numpy/reference/generated/numpy.atleast_1d.html

```ipython
In [14]: np.atleast_1d(5)
Out[14]: array([5])

In [15]: np.atleast_1d([5,6,7])
Out[15]: array([5, 6, 7])
```

Maybe this takes care of both issues?

PR #91, opened 2017-04-21
Schwefel20 tutorial and constrained udp end

darioizzo:


include/pagmo/algorithms/mbh.hpp
```diff
@@ -393,6 +393,16 @@ class mbh
     {
         return m_algorithm;
     }
+    /// Setter for the inner algorithm.
```
bluescarni: Isn't this superfluous? We have the getter above that already returns a non-const reference.

darioizzo: forgot to remove it .... I had added it as I wanted to replicate nlopt API for the local_optimizer, where the property is exposed in python as not read-only.

Then I changed my mind but forgot to remove this.

PR #92, opened 2017-04-24
Ipopt

bluescarni:


include/pagmo/algorithms/ipopt.hpp
```diff
@@ -0,0 +1,1179 @@
...
+ *   if the constraint tolerance is not explicitly set by the user, a solution deemed feasible by Ipopt is also
+ *   deemed feasible by pagmo (but the opposite is not necessarily true).
+ *
+ * \verbatim embed:rst:leading-asterisk
+ * .. warning::
+ *
+ *    A moved-from pagmo::ipopt is destructible and assignable. Any other operation will result
+ *    in undefined behaviour.
+ *
+ * .. note::
```
darioizzo: I thought we were using the syntax ``**NOTE**`` for doxygen .... are switching to this one? or its th rst embed instruction above that makes the diff?

include/pagmo/algorithms/ipopt.hpp
```diff
@@ -0,0 +1,1179 @@
...
+        m_integer_opts.clear();
+    }
+    /// Clear all numeric options.
+    void reset_numeric_options()
+    {
+        m_numeric_opts.clear();
+    }
+    /// Thread safety level.
+    /**
+     * According to the documentation, it is not safe to use Ipopt in a multithreaded environment.
```
darioizzo: According to Ipopt official documentation

include/pagmo/algorithms/ipopt.hpp
bluescarni: At one point I realised it's possible to embed sphinx syntax into doxygen with the construct above. It's a bit more cumbersome and, of course, looks like shit in the doxygen output, but renders much better in sphinx.

darioizzo: good to know ... we may as well then then change all **NOTE** to this format ...

PR #102, opened 2017-05-10
(WIP) Topology

bluescarni:
Just opening this early to check CI closely.


wardzin:
I don't mean to rush, but what's the current progress on the topology feature? Is it close to being done, or is it still far away from release?

bluescarni:
This branch implements some initial bits and pieces for the topology work, but it does not contain yet the critical part, which is the implementation of the migration within an archipelago.

There are a couple of reasons why this is taking long, but the main one is that we are essentially waiting to see if some funding materializes for the completion of the topology work. We should be able to clarify the situation soon (in a couple of weeks or so).

erl987:
Has there been any progress on this pull request? I guess migration within the archipelago is a pretty important feature of `pygmo` for practical usage of an archipelago?

It is not good that the documentation of `pygmo2` does not state at all that this feature is not implemented. Instead the docs of `pygmo` encourage users to not use it anymore. I spent several hours to understand that migration actually does not work in `pygmo2`.

darioizzo:
@erl987 Thanks for your note, yes you are perfectly right. Progress here has been slower than we anticipated and at the moment it is confusing to see migration assumed to be (implicitly) as part of the pagmo2 project. It is not.

We have now good ideas on how to implement migration in pagmo2 and hopefully we have recently received hints on some resources that will allow us to complete the PR.

bluescarni:
Closing in favor of #296. A first version of topology/migration will be part of the next pagmo release.

PR #107, opened 2017-05-18
fix cpp example bug in document.

maplewizard:
fix cpp example bug in document.


bluescarni:
@maplewizard Thanks! The travis build is failing due to issues unrelated to your changes, I am merging this.

PR #109, opened 2017-05-18
Iteration

bluescarni:
More bits of testing and a small doc fix.


bluescarni:
Ok I added the doc changes and bumped the version number for development as well.

Issue #112, opened 2017-05-25
Improve support to integers


darioizzo:
At the current status integer programming support is limited. The integer part of the decision vector can be passed as an argument to algorithms and the ``evolve()`` takes care to apply different operators on the float and the integer part. 

This way of doing things has obvious drawbacks and a better model would be to define an integer part of the decision vector in the problem.

 * [x] Introduce in the UDP interface a ``get_nix()``, that similarly to ``get_nic()``, ``get_nec()``, defines the integer part of the decision vector. (check that ``get_nix() <= get_nx()`` and that the bounds corresponding to the integer components are indeed integers)
 * [x] Modify the population constructor so that random decision vectors will have integers in their integer part (maybe modify the pagmo::random_decision_vector utility)
 * [x] Update the ``nsga2`` and ``sga`` algorithms removing the ``m_int_dim`` member and using the ``get_nix()`` from the ``pagmo::problem`` instead
 * [x] Update the algo list with the int information.

darioizzo:
#115 Is adressing this

darioizzo:
#115 is now merged into master ... minlp support will be available in 2.4

PR #115, opened 2017-05-28
Adding explicit MINLP support

darioizzo:
In reply to #112, this PR modifies the problem interface adding explicit support for MINLP type of problems. In order to do so the implemented logic is based on the following design drivers

 * All integers are represented as floats. 
 * pagmo only check is that the bounds of a problem are integers for the int part.
 * pagmo guarantees to generate integers when constructing a population, but does not prevent floats to be used in the decision vector where integers are expected.
 * Hessians and gradients are untouched. So the user can define them for the integer part and find
    the solution of the relaxed problem calling some NLP solver.
 * It is up to the algorithm to decide what to do with the integer part. By default, algorithms that are designed for continuous optimization will just treat it as a continuous variable, thus solving the relaxed problem.


doc/sphinx/docs/algorithm_list.rst
```diff
@@ -20,12 +21,12 @@ Differential Evolution (DE)                                :cpp:class:`pagmo::de
 Self-adaptive DE (jDE and iDE)                             :cpp:class:`pagmo::sade`                  :class:`pygmo.sade`                       S-U            
 Self-adaptive DE (de_1220 aka pDE)                         :cpp:class:`pagmo::de1220`                :class:`pygmo.de1220`                     S-U             
 Particle Swarm Optimization (PSO)                          :cpp:class:`pagmo::pso`                   :class:`pygmo.pso`                        S-U             
-(N+1)-ES Simple Evolutionary Algorithm                     :cpp:class:`pagmo::sea`                   :class:`pygmo.sea`                        S-U (sto)       
-Simple Genetic Algorithm                                   :cpp:class:`pagmo::sga`                   :class:`pygmo.sga`                        S-U (sto)       
+(N+1)-ES Simple Evolutionary Algorithm                     :cpp:class:`pagmo::sea`                   :class:`pygmo.sea`                        S-U   (sto)
```
bluescarni: Seems like an extra space slipped in here.

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
```diff
@@ -0,0 +1,171 @@
...
+    >>> pop = algo.evolve(pop)
+    >>> print("Best objective: ", pop.champion_f[0]) # doctest: +SKIP
+    Best objective:  134.065695174
+    >>> print("Best decision vector: ", pop.champion_x) # doctest: +SKIP
+    Best decision vector:  [ 0.4378605   0.33368365 -0.75844494 -1.          0.          0.        ]
+
+We found a feasible solution!
+
+.. note::
+   The solution strategy above is, in general, flawed in assuming the best solution of the relaxed problem is colse to the
```
bluescarni: "colse" -> "close"

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
```diff
@@ -0,0 +1,171 @@
...
+    >>> print("Best objective: ", pop.champion_f[0]) # doctest: +SKIP
+    Best objective:  134.065695174
+    >>> print("Best decision vector: ", pop.champion_x) # doctest: +SKIP
+    Best decision vector:  [ 0.4378605   0.33368365 -0.75844494 -1.          0.          0.        ]
+
+We found a feasible solution!
+
+.. note::
+   The solution strategy above is, in general, flawed in assuming the best solution of the relaxed problem is colse to the 
+   the full MINLP problem solution. More sophisticated techniques would instead search the combinatorial part more exhaustvely.
```
bluescarni: "exhaustively"

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
```diff
@@ -0,0 +1,171 @@
...
+   We used here this approach only to show how simple is, in pygmo, to define and solve the relaxed problem and
```
bluescarni: "How simple it is"

doc/sphinx/docs/python/tutorials/using_archipelago.rst
```diff
@@ -109,7 +109,7 @@ After inspection, let us now run the evolution.
     array([  4.07802374, -23.82020921,  -4.07802374]),
     array([  1.71396489, -25.90794514,  -1.71396489]),
     ...]
-    >>> archi.evolve()
+    >>> archi.evolve() #doctest: +SKIP
```
bluescarni: Why do this (and in a bunch of places below) need to be skipped as no output is being produced?

doc/sphinx/docs/python/tutorials/using_island.rst
```diff
@@ -37,15 +37,15 @@ on the opened thread.
 
 .. doctest::
 
-    >>> isl.evolve()
+    >>> isl.evolve()
```
bluescarni: Extra space.

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
                             + " while the upper bound has the smaller value " + std::to_string(ub[i]));
         }
     }
+    // 3 - checks the integer part
+    if (nix) {
+        auto nx = lb.size();
```
bluescarni: auto -> const auto

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
...
+        auto ncx = nx - nix;
+        if (nix > nx) {
```
bluescarni: I woud move the check one line above, before computing ncx. Also, ``const auto``.

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
...
+            pagmo_throw(std::invalid_argument, "The integer part cannot be larger than the bounds size");
+        }
+        for (decltype(ncx) i = ncx; i < nx; ++i) {
```
bluescarni: No need for ``decltype``. Here ``auto i = ncx`` works just fine.

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
...
+            if (lb[i] != std::floor(lb[i])) {
```
bluescarni: There is an ``std::trunc()`` function which IMO conveys the intent better than ``floor``. I would use it instead.

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
...
+                pagmo_throw(std::invalid_argument,
+                            "A lower bound of the integer part of the decision vector is: " + std::to_string(lb[i])
+                                + " and is not an integer.");
+            }
+            if (ub[i] != std::floor(ub[i])) {
```
bluescarni: Same here.

include/pagmo/problem.hpp
```diff
@@ -564,6 +606,26 @@ inline void check_problem_bounds(const std::pair<vector_double, vector_double> &
...
+                pagmo_throw(std::invalid_argument,
+                            "An upper bound of the integer part of the decision vector is: " + std::to_string(ub[i])
+                                + " and is not an integer.");
+            }
+        }
+    }
```
bluescarni: I think in these checks we are not handling the case of infinities right? I would put some checks for infinities (before the ``floor/trunc`` ones).

include/pagmo/problem.hpp
```diff
@@ -1058,6 +1139,7 @@ class problem
      * - the <tt>%gradient_sparsity()</tt> and <tt>%hessians_sparsity()</tt> methods of the UDP fail basic sanity checks
      *   (e.g., they return vectors with repeated indices, they contain indices exceeding the problem's dimensions,
      *   etc.).
+     * - the integer part of the problem is larger than the problem size.
```
bluescarni: I think we are missing some failure modes for ``check_problem_bounds()``: the integer bounds are not integers and the infinity check (if added above).

include/pagmo/problem.hpp
```diff
@@ -1066,9 +1148,17 @@ class problem
         : m_ptr(detail::make_unique<detail::prob_inner<uncvref_t<T>>>(std::forward<T>(x))), m_fevals(0u), m_gevals(0u),
           m_hevals(0u)
     {
+        // 0 - Integer part
+        auto tmp_size = ptr()->get_bounds().first.size();
```
bluescarni: const auto

include/pagmo/problem.hpp
```diff
@@ -1171,11 +1261,11 @@ class problem
...
-          m_nic(other.m_nic), m_c_tol(std::move(other.m_c_tol)), m_has_gradient(other.m_has_gradient),
-          m_has_gradient_sparsity(other.m_has_gradient_sparsity), m_has_hessians(other.m_has_hessians),
-          m_has_hessians_sparsity(other.m_has_hessians_sparsity), m_has_set_seed(other.m_has_set_seed),
-          m_name(std::move(other.m_name)), m_gs_dim(other.m_gs_dim), m_hs_dim(std::move(other.m_hs_dim)),
-          m_thread_safety(std::move(other.m_thread_safety))
+          m_nic(other.m_nic), m_nix(other.m_nix), m_c_tol(std::move(other.m_c_tol)),
+          m_has_gradient(other.m_has_gradient), m_has_gradient_sparsity(other.m_has_gradient_sparsity),
+          m_has_hessians(other.m_has_hessians), m_has_hessians_sparsity(other.m_has_hessians_sparsity),
+          m_has_set_seed(other.m_has_set_seed), m_name(std::move(other.m_name)), m_gs_dim(other.m_gs_dim),
+          m_hs_dim(std::move(other.m_hs_dim)), m_thread_safety(std::move(other.m_thread_safety))
```
bluescarni: Can you remove the ``std::move`` on the construction of ``m_hs_dim`` for consistency with the rest?

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+ * \f]
+ * And Hessians (sparse as only the diagonal is non-zero) are:
+ * \f[
+ * 	H_{ii}\left(x_1,\ldots,x_n\right) = 2 + 10 \cdot 4\pi^2 \cdot\cos\left( 2\pi \cdot x_i \right)
+ * \f]
+ */
+struct minlp_rastrigin {
+    /// Constructor from continuous and integer dimension
+    /**
+     * Constructs a MINLP Rastrigin problem
```
bluescarni: Missing period.

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        if (dim_c + dim_i < 1u) {
+            pagmo_throw(std::invalid_argument,
+                        "Minlp Rastrigin Function must have minimum 1 dimension, " + std::to_string(dim_c + dim_i)
+                            + " requested");
+        }
+    };
+    /// Fitness computation
+    /**
+     * Computes the fitness for this UDP
```
bluescarni: Missing period.

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+        for (decltype(n) i = 0u; i < n; ++i) {
+            f[0] += x[i] * x[i] - 10. * std::cos(omega * x[i]);
+        }
+        f[0] += 10. * static_cast<double>(n);
+        return f;
+    }
+
+    /// Box-bounds
+    /**
+     *
```
bluescarni: Extra line.

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+        sparsity_pattern hs;
+        auto n = m_dim_c + m_dim_i;
+        for (decltype(n) i = 0u; i < n; ++i) {
+            hs.push_back({i, i});
+        }
+        return {hs};
+    }
+    /// Problem name
+    /**
+     *
```
bluescarni: Extra lines.

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+     * @return a string containing the problem name
+     */
+    std::string get_name() const
+    {
+        return "MINLP Rastrigin Function";
+    }
+    /// Extra informations
+    /**
+     *
```
bluescarni: Extra line.

include/pagmo/problems/minlp_rastrigin.hpp
```diff
@@ -0,0 +1,248 @@
...
+     * @return a string containing extra informations on the problem
+     */
+    std::string get_extra_info() const
+    {
+        std::ostringstream ss;
```
bluescarni: Why go through a stringstream rather than just using string concatenation ``+``?

include/pagmo/utils/generic.hpp
```diff
@@ -117,37 +117,51 @@ inline double uniform_real_from_range(double lb, double ub, detail::random_engin
...
  * @endcode
  *
  * @param bounds an <tt>std::pair</tt> containing the bounds
  * @param r_engine a <tt>std::mt19937</tt> random engine
+ * @param nix size of the integer part
  *
  * @throws std::invalid_argument if:
  * - the bounds are not of equal length, they have zero size, they contain NaNs or infs,
- *   or \f$ \mathbf{ub} < \mathbf {lb}\f$,
+ *   \f$ \mathbf{ub} < \mathbf {lb}\f$, the integer part is larger than the bounds size or
```
bluescarni: Infinity check failure mode, if implemented.

include/pagmo/utils/generic.hpp
```diff
@@ -117,37 +117,51 @@ inline double uniform_real_from_range(double lb, double ub, detail::random_engin
...
+    detail::check_problem_bounds(bounds, nix);
+    auto nx = bounds.first.size();
+    vector_double retval(nx);
+    auto ncx = nx - nix;
 
-    for (decltype(dim) i = 0u; i < dim; ++i) {
+    for (decltype(ncx) i = 0u; i < ncx; ++i) {
         retval[i] = uniform_real_from_range(bounds.first[i], bounds.second[i], r_engine);
     }
+    if (nix > 0u) {
```
bluescarni: I don't think this is the right way to go.

First, we are hard-coding the integer type to ``int`` via ``std::uniform_int_distribution<>``. This effectively means restricting the range of the integer variables to 32 bits on modern platforms. This seems a rather important (and undocumented) implementation characteristic which may be limiting (?).

Secondly, the static casts are going to result in undefined behaviour if the floating point values are outside the ``int`` range.

What about instead generating numbers with a uniform real distribution in the specified range and calling something like ``std::round`` on them? I'd like to avoid going through a conversion to a C++ integer type, as that it's a shitcan to deal with and in the end we are always dealing with floating point values in the bounds, dvs etc. anyway.

include/pagmo/utils/generic.hpp
```diff
@@ -157,30 +171,35 @@ inline vector_double random_decision_vector(const std::pair<vector_double, vecto
...
  * @endcode
  *
  * @param lb a vector_double containing the lower bounds
  * @param ub a vector_double containing the upper bounds
  * @param r_engine a <tt>std::mt19937</tt> random engine
+ * @param nix size of the integer part
  *
  * @throws std::invalid_argument if:
- * - the bounds are not of equal length, they contain NaNs or infs, or \f$ \mathbf{ub} < \mathbf {lb}\f$,
+ * - the bounds are not of equal length, they have zero size, they contain NaNs or infs,
```
bluescarni: Missing failure modes.

pygmo/docstrings.cpp
```diff
@@ -2658,6 +2708,32 @@ See also the docs of the relevant C++ method :cpp:func:`pagmo::simulated_anneali
...
+are guaranteed to be integers within the specified (integer) bounds.
+
+Args:
+    lb (array-like object): the lower bounds
+    ub (array-like object): the upper bounds
+    nix (``int``): the integer size
+
+Raises:
+    ValueError: if *nix* is negative or greater than an implementation-defined value
+    ValueError: if *lb* and *ub* are malformed (unequal lenght, zero size, *nix* larger than len(lb) or bounds are not integers in their last *nix* components)
```
bluescarni: ``lenght``-> ``length``, missing failure mode, "len(lb)" should probably be in backticks as it's a fragment of code.

pygmo/test.py
```diff
@@ -844,6 +844,42 @@ def runTest(self):
...
+
+    def runTest(self):
+        from .core import random_decision_vector, set_global_rng_seed
+        set_global_rng_seed(42)
+        x = random_decision_vector(lb = [1.1,2.1,-3], ub = [2.1, 3.4,5], nix = 1)
+        self.assertTrue(int(x[-1]) == x[-1])
+        self.assertTrue(int(x[1]) != x[1])
+        set_global_rng_seed(42)
+        y = random_decision_vector(lb = [1.1,2.1,-3], ub = [2.1, 3.4,5], nix = 1)
+        self.assertTrue((x == y).all())
```
bluescarni: Would be good to have a couple of tests for failure modes as well.

tests/generic.cpp
```diff
@@ -71,16 +71,33 @@ BOOST_AUTO_TEST_CASE(random_decision_vector_test)
     BOOST_CHECK_THROW(random_decision_vector({{1, -inf}, {0, 32}}, r_engine), std::invalid_argument);
     BOOST_CHECK_THROW(random_decision_vector({{1, 2, 3}, {0, 3}}, r_engine), std::invalid_argument);
     BOOST_CHECK_THROW(random_decision_vector({{0, 2, 3}, {1, 4, nan}}, r_engine), std::invalid_argument);
+    BOOST_CHECK_THROW(random_decision_vector({{0, 2, nan}, {1, 4, 4}}, r_engine), std::invalid_argument);
```
bluescarni: Maybe add tests for the infinity checks?

bluescarni:
@darioizzo looks good. The only real issue I have found is the one about generating random decision vectors (see comments).

doc/sphinx/docs/python/tutorials/using_archipelago.rst
darioizzo: It does not need to be skipped, I choose to skip it as it makes the test much longer.

include/pagmo/problems/minlp_rastrigin.hpp
darioizzo: In general I prefer the use of stringstream syntax over the string concatenation as it reads more clearly (after format) and allows for the immediate use of our printing functionalities. In this case, though, there is no need and its here for consistency with my own way of doing it :)

include/pagmo/problem.hpp
darioizzo: fatto ma perche qui si e nella linea successiva non e importante?

include/pagmo/problem.hpp
darioizzo: ah ok

include/pagmo/problem.hpp
darioizzo: l abitudine :)

include/pagmo/problem.hpp
darioizzo: You probabloy know already, but be aware they are not exactly the same.

trunc: -0.9 -> 0
floor: -0.9 -> -1

In this case it does not matter though ....

include/pagmo/problem.hpp
darioizzo: In my mind we did not have to check for infinite values as we do allow infinte bounds for floats and our integers are just floats.  But it would also make sense to forbid infinite values in the case of integers .... breaking though the symmetry int / floats.

include/pagmo/problem.hpp
darioizzo: To me that was all included in the etc. above :) But ok will explicit it

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
darioizzo: done

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
darioizzo: done

doc/sphinx/docs/python/tutorials/coding_udp_minlp.rst
darioizzo: done

include/pagmo/problem.hpp
darioizzo: done

include/pagmo/problem.hpp
darioizzo: I leave it as is for this review, let me know

include/pagmo/utils/generic.hpp
darioizzo: Ok I then have to use floor and add 1 to the upper bound as to ensure the obtained distribution from int is uniform in the bounds.

pygmo/test.py
darioizzo: those are tested in the C++ part. I usually only test the boost interface in test.py

tests/generic.cpp
darioizzo: So I think there is some confusion on how we deal with infinites. 

Status quo:

 * It is allowed to have infinite bounds on the problem.
 * It is not allowed to draw at random a number if the bounds are infinite.

These cases are already tested on the c++ side.

With the addition of the integer part I would not change this behaviour and keep the possibility to have infinite bounds on the problem (also for integers), while preventing to draw random decision vectors from infinite bounds.

The reason for this seemingly asymmetric behaviour is that
 * It makes sense to signal to the optimizer `algo.evolve()` that the problem is unbounded.
 * It makes little sense, if the bounds are infinite, to construct a decision vector drawing randomly from it, better to have the user explicitly init the pop in that case.

include/pagmo/utils/generic.hpp
darioizzo: added (except infinites)

include/pagmo/utils/generic.hpp
darioizzo: done

include/pagmo/utils/generic.hpp
darioizzo: done

include/pagmo/utils/generic.hpp
```diff
@@ -117,37 +117,53 @@ inline double uniform_real_from_range(double lb, double ub, detail::random_engin
...
+    detail::check_problem_bounds(bounds, nix);
+    auto nx = bounds.first.size();
+    vector_double retval(nx);
+    auto ncx = nx - nix;
 
-    for (decltype(dim) i = 0u; i < dim; ++i) {
+    for (decltype(ncx) i = 0u; i < ncx; ++i) {
         retval[i] = uniform_real_from_range(bounds.first[i], bounds.second[i], r_engine);
     }
+    if (nix > 0u) {
```
bluescarni: Two notes:

- I think you can skip the ``if (nix > 0u)`` because you have already the ``for`` cycle checking for the same condition right?
- How about using ``std::nextafter`` for having a closed interval. See: http://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution/uniform_real_distribution

> To create a distribution over the closed interval [a,b], std::nextafter(b, std::numeric_limits<RealType>::max()) may be used as the second parameter.

include/pagmo/utils/generic.hpp
```diff
@@ -117,37 +117,53 @@ inline double uniform_real_from_range(double lb, double ub, detail::random_engin
...
+        for (decltype(nx) i = ncx; i < nx; ++i) {
```
bluescarni: ``auto i = ncx``

tests/generic.cpp
bluescarni: Right, I forgot that the random vector stuff errors out with infinities.

pygmo/test.py
bluescarni: I would consider exception translation part of the boost interface.

PR #119, opened 2017-06-01
pop has not get_name problem

ahcheriet:
class pop has no get_name member


ahcheriet:
Problem has no get_name member

darioizzo:
@ahcheriet Indeed thanks for your PR! .... this was being addressed also in #118 so I will close it.

PR #121, opened 2017-06-12
Small fixes

bluescarni:
It was reported on the gitter chat that the docs of some metas are not 100% accurate - specifically, some constructors do not mention that the metas can be constructed also with a problem/algorithm instance (and not only with a UDP/UDA).

I also fixed the usage of anonymous hyperlinks in the sphinx docs (https://stackoverflow.com/questions/5464627/how-to-have-same-text-in-two-links-with-restructured-text) and fixed some random stuff I rant into while cleaning up. Unfortunately, there's a bunch of clang-format noise as well.

### DIFF
--- a/doc/sphinx/changelog.rst
+++ b/doc/sphinx/changelog.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+2.5 (unreleased)
+----------------
+
+Fix
+~~~
+
+- Various documentation fixes and improvements.
+
 2.4 (2017-06-09)
 ----------------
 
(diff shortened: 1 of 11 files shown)


doc/sphinx/changelog.rst
```diff
@@ -10,57 +18,57 @@ New
 - Initial release of the pagmo/pygmo C++ software-development kit (SDK). The purpose of the SDK is to make the process
   of writing C++ extensions for pagmo/pygmo as easy as possible. The SDK is a beta-quality feature at this time,
   and it is lightly documented - no tutorials are available yet. Please come to the gitter channel and ask there if you are interested
-  in it (`#110 <https://github.com/esa/pagmo2/pull/110>`_).
-
-- Improve support for integer and mixed integer optimization (`#115 <https://github.com/esa/pagmo2/pull/115>`_).
+  in it (`#110 <https://github.com/esa/pagmo2/pull/110>`__).
```
darioizzo: Can you add to this PR something like following text: "The problem integer part is no longer available as argument when constructing algorithms such as pagmo::sga and pagmo::nsga2, it is must instead be specified implementing a method get_nix on the UDP". 

Since this breaks previous code it may be good to have it logged.

bluescarni:
ok?

PR #128, opened 2017-08-12
CI iteration

bluescarni:


bluescarni:
@darioizzo a new version of breathe came out that fixes the issues with sphinx 1.6, so now in the CI we are using the latest versions of all doc-related packages. Unfortunately, there seem to be some issues with the Python 2.7 doc building, but at this point I am for just not checking the docs in the Python 2 build and just assume Python 3.x in the docs.

bluescarni:
@darioizzo there seemed to be an issue in the sga docs, which referred to the ``int_dim`` construction parameter which AFAIR was removed in the minlp problem rework some time ago.

PR #136, opened 2017-11-02
More CI fixes.

bluescarni:


bluescarni:
Merging this as a temp workaround until codecov is fixed.

Issue #138, opened 2017-11-14
Missing documentation of get_nix() on the python side


bluescarni:
The problem's ``get_nix()`` method is not documented on the Python side.

bluescarni:
Fixed in #139.

Issue #137, opened 2017-11-14
NSGA II not dealing correctly with negative integers


darioizzo:
The algorithm nsga ii on the line:
https://github.com/esa/pagmo2/blob/master/include/pagmo/algorithms/nsga2.hpp#L539

mutates an integer decision vector component within its bounds, but it casts them to unsigned int, thus creating overflows or UBs when the bounds contain negative values (which is allowed in pagmo)

darioizzo:
https://github.com/esa/pagmo2/blob/master/include/pagmo/algorithms/nsga2.hpp#L537 this should also be corrected to <0 not <=0

bluescarni:
@darioizzo we can close this I imagine?

darioizzo:
yeps.

PR #139, opened 2017-11-15
[CI skip] Nsgaii fix

darioizzo:
A fix for #138 and #137.

Essentially using the pagmo utility ``generate_random_vector`` to avoid the conversion madness from float to int.

### DIFF
--- a/include/pagmo/algorithms/nsga2.hpp
+++ b/include/pagmo/algorithms/nsga2.hpp
@@ -44,7 +44,7 @@ see https://www.gnu.org/licenses/. */
 #include <pagmo/problems/decompose.hpp>
 #include <pagmo/rng.hpp>
 #include <pagmo/utils/multi_objective.hpp> // crowding_distance, etc..
-#include <pagmo/utils/generic.hpp> // radnom_decision_vector
+#include <pagmo/utils/generic.hpp> // uniform_real_from_range
 
 namespace pagmo
 {
@@ -537,10 +537,10 @@ private:
         for (decltype(D) j = Dc; j < D; ++j) {
             if (drng(m_e) < m_m) {
                 // We need to draw a random integer in [lb, ub]. Since these are floats we
-                // cannot use integer distributions without risking overflows, hence we use the 
-                // pagmo utility which takes care of it random_decision_vector
-                auto mutated = random_decision_vector({{lb[j]},{ub[j]}}, m_e, 1u);
-                child[j] = mutated[0];
+                // cannot use integer distributions without risking overflows, hence we use a real
+                // distribution
+                auto mutated = std::floor(uniform_real_from_range(lb[j], ub[j] + 1, m_e));
+                child[j] = mutated;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


pygmo/docstrings.cpp
```diff
@@ -546,6 +548,11 @@ Integer dimension of the problem.
 
 This method will return :math:`n_{ix}`, the integer dimension of the problem.
 
+The optional ``get_nix()`` method of the UDP must return the number of equality constraints as an ``int``.
```
bluescarni: "number of equality constraints" --> "problem's integer dimension"

include/pagmo/algorithms/nsga2.hpp
```diff
@@ -534,10 +535,12 @@ class nsga2
...
         for (decltype(D) j = Dc; j < D; ++j) {
-            if (drng(m_e) <= m_m) {
-                std::uniform_int_distribution<vector_double::size_type> ra_num(
-                    static_cast<vector_double::size_type>(lb[j]), static_cast<vector_double::size_type>(ub[j]));
-                child[j] = static_cast<double>(ra_num(m_e));
+            if (drng(m_e) < m_m) {
+                // We need to draw a random integer in [lb, ub]. Since these are floats we
+                // cannot use integer distributions without risking overflows, hence we use the 
+                // pagmo utility which takes care of it random_decision_vector
+                auto mutated = random_decision_vector({{lb[j]},{ub[j]}}, m_e, 1u);
```
bluescarni: This will do a memory allocation at every invocation of ``random_decision_vector()``. Are we ok with the performance implications of this? Is this code gonna be executed very frequently?

A solutio would be to have an additiona overload of ``random_decision_vector()`` that takes ``mutated`` (the return value) as a first parameter by preference.

darioizzo: thought about it, indeed it gets called a lot, but not sure would be a bottle neck ....

darioizzo: I can just code it on the fly without invoking the utility

darioizzo: I can just put the code
```
        if (!std::isfinite(ub + 1.)) {
            pagmo_throw(std::invalid_argument,
                        "The upper bound of the integer part is too large for the decision vector index "
                            + std::to_string(i));
        }

        mutated = uniform_real_from_range(lb, ub + 1, m_e);
```

bluescarni: You need the floor as well right? And isn't ``uniform_real_from_range()`` already checking for infinity in the bounds?

darioizzo: the floor is there to get an int out of a float and ensure the drawn distribution is uniform from lb to ub. We use the same trick in ``random_decision_vector``

The infinity bounds are indeed checked in ``uniform_real_from_range`` so maybe we should also remove that check in ``random_decision_vector``?

bluescarni: Yep it looks like the check should be removed in ``random_decision_vector()`` as well.

bluescarni:
@darioizzo Looks good to me. I'd remove the commented code once the CI runs, and also remove the comment that starts with ``NOTE: to pursue this approach we need to...`` as it no longer applies.

After that we can merge I think.

PR #142, opened 2017-12-01
Exponential Natural Evolution Strategies

darioizzo:
The algorithm xNES is closely related to CMA-ES as it uses and adapts a gaussian multivariate distribution. Its update rules are far more elegant at the expense of performances (with cma-es having distinct advantages on  the usual problems). The c++ implementation in pagmo is taken from the paper  (http://people.idsia.ch/~juergen/xNES2010gecco.pdf) with the guiding help of the pybrain implementation here (https://github.com/pybrain/pybrain/blob/master/pybrain/optimization/distributionbased/xnes.py)

We introduced two changes to the original algorithm in order to simplify its use for the generic user.
 *    1 - when a decision vector is sampled outside the problem bounds it may be forced back in.
 *    2 - the initial covariance matrix depends on the bounds width so that heterogenously scaled variables are not a problem.


darioizzo:
Green at last!! good to go for me.

include/pagmo/algorithms/cmaes.hpp
```diff
@@ -83,8 +83,23 @@ class cmaes
 {
 public:
     /// Single entry of the log (gen, fevals, best, dx, df, sigma)
-    typedef std::tuple<unsigned int, unsigned long long, double, double, double, double> log_line_type;
-    /// The log
+    /// Single data line for the algorithm's log.
```
bluescarni: Seems to be two titles here.

include/pagmo/algorithms/cmaes.hpp
```diff
@@ -83,8 +83,23 @@ class cmaes
...
+    /**
+     * A log data line is a tuple consisting of:
+     * - the generation number,
+     * - the number of function evaluations
+     * - the best fitness vector so far,
+     * - the population flatness evaluated as the distance between the decisions vector of the best and of the worst
+     * individual,
+     * - the population flatness evaluated as the distance between the fitness of the best and of the worst individual.
+     */ typedef std::tuple<unsigned int, unsigned long long, double, double, double, double>
```
bluescarni: This should probably go onto the next line. Kinda surprised clang-format did not pick it up.

include/pagmo/algorithms/cmaes.hpp
```diff
@@ -83,8 +83,23 @@ class cmaes
...
+        log_line_type;
+    /// Log type.
+    /**
+     * The algorithm log is a collection of nlopt::log_line_type data lines, stored in chronological order
```
bluescarni: nlopt?

include/pagmo/algorithms/xnes.hpp
```diff
@@ -0,0 +1,590 @@
...
+ *
+ *    Glasmachers, T., Schaul, T., Yi, S., Wierstra, D., & Schmidhuber, J. (2010, July). Exponential natural
+ *    evolution strategies. In Proceedings of the 12th annual conference on Genetic and evolutionary computation (pp.
+ *    393-400). ACM.
+ * \endverbatim
+ */
+class xnes
+{
+public:
+    /// Single entry of the log (gen, fevals, best, dx, df, sigma)
```
bluescarni: Double title here as well.

include/pagmo/algorithms/xnes.hpp
```diff
@@ -0,0 +1,590 @@
...
+     * - the number of function evaluations
+     * - the best fitness vector so far,
+     * - the population flatness evaluated as the distance between the decisions vector of the best and of the worst
+     *   individual,
+     * - the population flatness evaluated as the distance between the fitness of the best and of the worst individual.
+     */
+    typedef std::tuple<unsigned int, unsigned long long, double, double, double, double> log_line_type;
+    /// Log type.
+    /**
+     * The algorithm log is a collection of nlopt::log_line_type data lines, stored in chronological order
```
bluescarni: nlopt

include/pagmo/algorithms/xnes.hpp
```diff
@@ -0,0 +1,590 @@
...
+     * One of the optional methods of any user-defined algorithm (UDA).
+     *
+     * @return a string containing extra informations on the algorithm
+     */
+    std::string get_extra_info() const
+    {
+        std::ostringstream ss;
+        stream(ss, "\tGenerations: ", m_gen);
+        stream(ss, "\n\teta_mu: ");
+        if (m_eta_mu == -1)
```
bluescarni: Would be better to put brackets around all these ``if``/``else``.

PR #144, opened 2017-12-06
A few CI fixes/improvements

bluescarni:


bluescarni:
I stopped some builds to shorten the wait, but it seems like the manylinux builds are working again. I updated some deps and moved the coverage build out of the allowed_failures section.

PR #164, opened 2018-04-11
Small CI/build system bits.

bluescarni:


include/pagmo/algorithms/pso_gen.hpp
```diff
@@ -96,7 +98,7 @@ const int pso_gen_statics<T>::vonNeumann_neighb_diff[4][2] = {{-1, 0}, {1, 0}, {
  *
  * \endverbatim
  */
-class pso_gen : public pso_gen_statics<>
```
darioizzo: preferisci non derivare ed accedere tramite namespace detail? Era cosi, poi lo avevo cambiato, mi sembrava piu corto :)

include/pagmo/serialization.hpp
```diff
@@ -29,13 +29,51 @@ see https://www.gnu.org/licenses/. */
...
+#endif
+
+// LLVM 3.7.
+#if __clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 7)
+
+#pragma GCC diagnostic ignored "-Wexceptions"
+
+#endif
+
+#endif
```
darioizzo: a che serve sta roba?

darioizzo: GO ahead and merge

include/pagmo/algorithms/pso_gen.hpp
bluescarni: There was a public inheritance from a class which is not even showing up in the sphinx docs, and which really looks like an internal implementation detail. Perhaps it could also be shared between the two pso implementations, Don't Repeat Yourself and all that jazz :)

include/pagmo/serialization.hpp
bluescarni: Selectively disable some clang warnings generated by the cereal headers.

It turns out that Apple, continuing their war against their users, hijack the clang version macro with Xcode's version string. So, let's say that you need to disable the "-Wexceptions" error warning, which is available since clang 3.7. You need to figure out from which version of Xcode clang 3.7 is being used internally, and write two sets of macros depending on whether you are using the "normal" clang or appleclang. Fun stuff.

PR #167, opened 2018-04-28
Modified plot_non_dominated_fronts so it does not create multiple axes on Jupyter notebooks

KishManani:
I've changed the code so that it does not use both plt.show() and plt.gca() which results in two figures being plotted on a Jupyter notebook. Instead it creates a figure first and returns the axes object. Now only a single figure is created when using plot_non_dominated_fronts() on a Jupyter notebook as well as other IDEs like PyCharm.


KishManani:
The CI failing appears to be unrelated to changes I've made to the plotting. What would I need to do to get this merged? Thanks. @darioizzo

darioizzo:
cool can you remove the file  pygmo/__pycache__/__init__.cpython-36.pyc ? It may have slipped in the commit

darioizzo:
Thanks! for this PR! It good to go as far as I am concerned. @bluescarni I suggest to fix the CI and then merge this, if you are ok with it.

bluescarni:
conda updated to the latest Boost version 1.67, which introduces a naming change in the Boost.Python library. I have opened #169 with a tentative fix.

bluescarni:
@KishManani could you please merge master into this PR? That should take care of the CI issues.

PR #186, opened 2018-07-05
UDIs in Python must now return the algorithm used for the evolution i…

bluescarni:
…n addition to the evolved population.


bluescarni:
@darioizzo can you check if this solves the xnes issue?

darioizzo:
on it

Issue #171, opened 2018-05-09
Is it possible to kill the archipelago processes after completion?


bsugerman:
I'm running pygmo in Anaconda on windows. After I'm done running an archipelago, all the python processes that it opened remain idle and only die if I quit the python interpreter. Is there a hidden way to shut the archipelago processes down once I'm done using them? Or can this be added as a feature, i.e. `archi.close()`?

bsugerman:
I think that exposing the multiprocessing `close()` and `join()` is a very useful feature. I'm finding extremely erratic behavior in my archipelago behavior because so many python threads are left running after the archipelago has finished. Although I suppose I should ask whether the current version uses `multiprocessing` or another method to run islands in parallel.

bluescarni:
@bsugerman Sorry I initially missed the first part of our reply. What do you mean by "extremely erratic behaviour" exactly?

By default on Windows the current pygmo version uses ``multiprocessing`` if the algorithm or the problem are implemented in Python. If both the algorithm and the problem are implemented in C++, it will use multithreading.

bluescarni:
@bsugerman There's an implementation of this functionality up at #187.

bluescarni:
#187 has now been merged to master (which will soon become pagmo 2.8). I'll close this issue report.

PR #187, opened 2018-07-05
Add the possibility of shutting down explicitly the process pool in mp_island

bluescarni:
Also, start using intersphinx.


bluescarni:
This aims at addressing #171.

pygmo/_island_test.py
```diff
@@ -275,10 +275,17 @@ def run_basic_tests(self):
         from . import mp_island
         from copy import copy, deepcopy
         from pickle import dumps, loads
+        # Try shutting down a few times.
```
darioizzo: Maybe just add some explanation on why ... it looks weird to have these lines repeated one may think its just a typo. (same below.

PR #188, opened 2018-07-06
Adds CEC2014 problems

brenoarosa:
Hi guys,

In this fork I added CEC2014 problems heavily inspired by the existing CEC2013 UDP.
The original competition files and paper can be found at http://www.ntu.edu.sg/home/EPNSugan/index_files/CEC2014/CEC2014.htm

As with CEC2013, additional data files were translated to pagmo/detail/cec2014_data.hpp which is a huge file and slows down quite a bit the compilation.

This is still a Work In Progress as it still miss tests and documentation update.
Unfortunately, I'm not well versed in C++ and I'm having trouble running the tests. I would appreciate any help in this concern.

### DIFF
--- a/include/pagmo/problems/cec2014.hpp
+++ b/include/pagmo/problems/cec2014.hpp
@@ -33,6 +33,29 @@ typedef std::vector<double> vector_double;
 
 namespace pagmo
 {
+/// The CEC 2014 problems: Real-Parameter Single Objective Optimization Competition
+/**
+ *
+ * The 30 problems of the competition on real-parameter single objective optimization problems that
+ * was organized for the 2014 IEEE Congress on Evolutionary Computation.
+ *
+ * \verbatim embed:rst:leading-asterisk
+ * .. note::
+ *
+ *    The code for these UDAs is adapted from the original C code distributed during the competition and
+ *    linked below.
+ *
+ * .. note::
+ *
+ *    All problems are box-bounded, continuous, single objective problems.
+ *
+ * .. seealso:
+ *
+ *    http://www.ntu.edu.sg/home/EPNSugan/index_files/CEC2014/CEC2014.htm
+ *    http://web.mysites.ntu.edu.sg/epnsugan/PublicSite/Shared%20Documents/CEC-2014/cec14-c-code.zip
+ *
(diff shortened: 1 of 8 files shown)


include/pagmo/problems/cec2014.hpp
```diff
@@ -0,0 +1,1449 @@
...
+        nx = dim;
+
+        if (!(dim == 2u || dim == 10u || dim == 20u || dim == 30u || dim == 50u || dim == 100u)) {
+            pagmo_throw(std::invalid_argument, "Error: CEC2014 Test functions are only defined for dimensions "
+                                               "2,10,20,30,50,100, a dimension of "
+                                                   + std::to_string(dim) + " was detected.");
+        }
+        if (prob_id < 1u || prob_id > 30u) {
+            pagmo_throw(std::invalid_argument,
+                        "Error: CEC2013 Test functions are only defined for prob_id in [1, 28], a prob_id of "
```
bluescarni: Typo 2013 -> 2014.

pygmo/test.py
```diff
@@ -994,6 +994,20 @@ def runTest(self):
         udp = cec2013(prob_id=3, dim=10)
 
 
+class cec2014_test_case(_ut.TestCase):
+    """Test case for the UDP cec2013
```
bluescarni: Typo 2013 -> 2014.

bluescarni:
@brenoarosa Thanks for the PR!

I left a couple of comments about some minor typos.

It looks like you forgot to add to the git tree the .cpp test file for the new UDP, which is making some build in the CI fail. Could you add it so that maybe @darioizzo (which originally wrote the CEC2013 problem) can chime in and take a look?

darioizzo:
Breno,
thanks for your work on the PR. There are several small problems on the code but I am not going to list them here as anyway there also is a nasty problem related to the size of the static data you allocate that requires us to make some decisions.

So I have taken your code and will soon open a PR on my name where you can also follow to see the small changes (as a curiosity the test was not going as you construct problems of dimensions 2 but some of them do not exist hence the throw). FYI you can see the exact text of the throw if you run the test directly and not via ```make test```)

PR #189, opened 2018-07-09
CEC2014

darioizzo:
The cec2014 problems from @brenoarosa.
Trying to solve issues of the original pull request #188 from the user @brenoarosa


darioizzo:
@brenoarosa I changed a bunch of stuff to improve code quality. Since the test do not check the correctness of the actual values computed by the obj fun, could you quickly check they are indeed correct?

In essence I would need you to call the various functions on one point and check that the value computed is correct (one could actually put this in the test). 

Do you think you can do it? Or send me the list of chromosomes, objfunctions to test so that I can add it to the tests? (in which case make sure to send all digits).

brenoarosa:
Sure, the original code came with a test case, I'm porting it and will send a PR.

darioizzo:
@brenoarosa Ok thanks will put this in the tests. But you obtained the value from the code in pagmo, did you check the results are the same as the original C code?

darioizzo:
@bluescarni 
We still have failures on py35 appveyor (some conda problem with the issued update commands) and on the osx builds as (probably compilation of core times out?)

bluescarni:
Merging this for now, will try to sort out the CI/OSX woes in a separate PR.

brenoarosa:
Hi, sorry for the delay.
Yes, the values matches the test provided in the original code.
I used this values to validate the changes that I made.

PR #192, opened 2018-07-12
Add missing import (`setw`)

jakobj:
This PR adds an include that I need to add in order to compile on my machine.


bluescarni:
Thank you!

Issue #180, opened 2018-06-15
Tracking evolution for nlopt?


emmae0:
Hello,

I am using nlopt. I know that algo.extract(pg.nlopt).get_log() gives you a log with the fitness evaluations made, but is there anywhere I can find the input values (aka the x values) evaluated to find those fitness evaluations?

Thanks!

darioizzo:
i am afraid not .... not sure whether the nlopt library stores those values
at all ...

bluescarni:
I have added to pygmo a decorator meta-problem, that allows you (amongst other things) to modify a problem so that it logs the decision vectors that have been passed to it. See for instance the example in the documentation:

https://esa.github.io/pagmo2/docs/python/tutorials/udp_meta_decorator.html#logging-fitness-evaluations

This should allow you to log the information you need (not only with nlopt, but with any other algorithm).

I am closing the report for now, please open a new one if you find issues with the decorator problem approach.

PR #195, opened 2018-07-17
Decorator problem

bluescarni:
Opening this early for feedback.

This PR adds a python meta-problem that allows to add callbacks to the functions of the public API of a UDP. Minimal example:

```ipython
In [1]: import pygmo as pg

In [2]: def fitness_cb(self, dv):
   ...:     print(dv)
   ...:     

In [3]: p = pg.problem(pg.problem_callback(pg.rosenbrock(), fitness_callback=fitness_cb))

In [4]: p
Out[4]: 
Problem name: Multidimensional Rosenbrock Function [with callbacks]
        Global dimension:                       2
        Integer dimension:                      0
        Fitness dimension:                      1
        Number of objectives:                   1
        Equality constraints dimension:         0
        Inequality constraints dimension:       0
        Lower bounds: [-5, -5]
        Upper bounds: [10, 10]

        Has gradient: true
        User implemented gradient sparsity: false
        Expected gradients: 2
        Has hessians: false
        User implemented hessians sparsity: false

        Fitness evaluations: 0
        Gradient evaluations: 0

        Thread safety: none

Extra info:
        Registered callbacks:
                fitness

In [5]: p.fitness([1,2])
[1. 2.]
Out[5]: array([100.])

In [6]:
```

Here we are adding a callback to the fitness function that prints to screen the value of the input decision vector.

The callbacks must have the same signature as the original methods. For instance, in this case we are adding a callback to the fitness function, which takes as input parameters the ``self`` object and the decision vector ``dv``. Note that the ``self`` parameter that appears in the callback is not the original UDP (rosenbrock, in this case) but the ``problem_callback`` instance wrapping the UDP.

This is an attempt to provide a mechanism to solve #180. Using callbacks, it is possible for instance to log all the ``dv``s evaluated by a problem, with a fitness callback like this:

```ipython
In [13]: def fitness_log_cb(self, dv):
    ...:     if not hasattr(self, "dv_log"):
    ...:         self.dv_log = []
    ...:     self.dv_log.append(dv)
    ...:     

In [14]: p = pg.problem(pg.problem_callback(pg.rosenbrock(), fitness_callback=fitness_log_cb))

In [15]: p.fitness([1,2])
Out[15]: array([100.])

In [16]: p.fitness([3,4])
Out[16]: array([2504.])

In [17]: p.extract(pg.problem_callback).dv_log
Out[17]: [array([1., 2.]), array([3., 4.])]

In [18]:
```

Comments?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,4 @@
         "doc/doxygen/xml/": true,
         "doc/sphinx/_build/": true
     },
-    "python.linting.pylintEnabled": false
 }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ install:
 # - if [%BUILD_TYPE%]==[MSVC_64_Debug] conda clean --all -y
 # - if [%BUILD_TYPE%]==[MSVC_64_Debug] conda update conda -y
 - if [%BUILD_TYPE%]==[MSVC_64_Debug] conda update --all -y
-- if [%BUILD_TYPE%]==[MSVC_64_Debug] conda create -y --name pagmo python=3.6 cmake boost boost-cpp eigen nlopt
+- if [%BUILD_TYPE%]==[MSVC_64_Debug] conda create -y --name pagmo python=3.6 cmake boost boost-cpp eigen nlopt numpy cloudpickle ipyparallel
 # NOTE: need to use "call" because otherwise it won't work within an if block.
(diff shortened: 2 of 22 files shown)


darioizzo:
* [] Provide python tests
 * [] Some docs?

bluescarni:
@darioizzo pending CI, this should now be ready if you want to give it another look.

PR #197, opened 2018-07-18
A round of new debug compiler flags and related fixes

bluescarni:


darioizzo:
whats with the override they were just all wrongly there?

bluescarni:
It's just a suggestion from the compiler. Those are all virtual methods being re-implemented in a derived class so it's better to signal it with the ``override`` keyword. It has no practical effect.

PR #205, opened 2018-08-04
Fork island

bluescarni:
An island that uses the ``fork()`` system call on POSIX systems to run the optimisation in a separate process. Useful as a replacement for ``thread_island`` for legacy problems with thread-unsafe global state.


darioizzo:
This also contains commits that fix some small docs inconsistencies (i.e. PSO), some small hypervolume code improvements (i.e. final keword use), and headers inconsistencies (i.e. eigen etc..)

PR #206, opened 2018-08-07
Add inline specifier to a function defined in header

jslee02:
This PR fixes the multiple definition errors.


bluescarni:
Thank you!

PR #219, opened 2018-09-07
Extraction via object

bluescarni:
This PR adds support for extracting Python user-defined entities without knowing their exact type.

For example, given the Python UDA

```python
class algo:
   def evolve(self, pop):
      return pop
```

one can now do:

```python
>>> a = pg.algorithm(algo())
>>> a.extract(object)
<__main__.algo at 0x7f8e478c04e0>
```

whereas previously one would need to write, instead,

```python
>>> a = pg.algorithm(algo())
>>> a.extract(algo)
<__main__.algo at 0x7f8e478c04e0>
```

This second syntax still works.

For exposed C++ user-defined classes, nothing changes (i.e., the exact correct type must still be provided for a successful extraction).

Two reasons why this might be useful.

First, working with pythonic classes is now slightly easier, as one does not need to remember the exact type for extraction.

Second, this can be useful to work around extraction issues when working in an interactive session.

Right now, if one defines a new UDP in an interactive session and then does the following

```python
In [1]: import pygmo as pg

In [2]: class prob:
   ...:     def fitness(self, x):
   ...:         return [x[0]]
   ...:     def get_bounds(self):
   ...:         return ([0],[1])
   ...:     

In [3]: isl = pg.island(prob=prob(), algo=pg.de(100), size=20)

In [4]: isl.evolve();isl.wait_check()
```

then, unexpectedly,

```python
In [5]: isl.get_population().problem.extract(prob) is None
True
```

That is, after evolution has been dispatched to an external process, the extraction of the problem via its original type fails. This happens because of various internal details of how Python serialisation works (specifically, the ``prob`` class that gets defined in the ``__main__`` module of the remote process is not of the same type as the original ``prob`` class). This may or may not be fixed in the future.

With ``object``-based extraction, we can now do instead

```python
In [6]: isl.get_population().problem.extract(object)
<__main__.prob at 0x7f48585d2b38>
```

and this will now work as intended.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ if((NOT PAGMO_BUILD_PAGMO) AND (NOT PAGMO_BUILD_PYGMO))
 endif()
 
 # Main pagmo/pygmo project version.
-set(PAGMO_PROJECT_VERSION 2.9)
+set(PAGMO_PROJECT_VERSION 2.10)
 
 if(PAGMO_BUILD_PAGMO)
     # Initial setup of a pagmo build.
--- a/doc/sphinx/changelog.rst
+++ b/doc/sphinx/changelog.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+2.10 (unreleased)
+-----------------
+
+Fix
+~~~
+
+- Tentative fix for a pygmo build failure in Cygwin (`#219 <https://github.com/esa/pagmo2/pull/219>`__).
+
+- Various documentation fixes and enhancements (`#217 <https://github.com/esa/pagmo2/pull/217>`__, `#218 <https://github.com/esa/pagmo2/pull/218>`__).
+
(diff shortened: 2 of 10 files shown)


bluescarni:
Staring at the code for a bit, I think we can even get to the point where, in Python, we can just avoid passing any argument to extract also for C++ entities (at the price of some overhead in the extract operation).

The basic idea is that we keep around a vector of ``std::function`` implementing the extraction for the various  exposed C++ types. Each time we expose a new prob/algo, we add an element to this vector implementing the extraction for that specific type. Then we write an ``extract()`` wrapper that iterates over the vector of functions, trying them one by one. Only one of these functions (the correct one) will return a non-null pointer during extraction.

I think this should work, but I am not gonna try to do it right now. I think it can be implemented in a backwards compatible way, so I'll go ahead and merge this for the time being.

PR #220, opened 2018-09-07
decompose_objectives docs update

bertji:
Added note on the lack of impact ref_point on the weighted method.
Added two additional examples showing the new note as well as an alternative example for tchebycheff method.


bluescarni:
@bertji thank you! It looks good to me, pinging @darioizzo as well (as he wrote the code for the ``decompose_objectives()`` function).

Issue #216, opened 2018-09-05
pagmo_export.cmake not found in Cygwin installation


astrophysaxist:
Hi, I've compiled pagmo2 successfully from source with Cygwin. I was able to test compiling the getting_started.cpp just fine. Now, I'm looking to install pygmo from source, but I've hit a snag I'm not sure how to troubleshoot. After updating the configuration of my cmake to build only pygmo, I get an error:

```
$ /usr/local/bin/cmake ..
-- System name: CYGWIN
-- The C++ compiler ID is: GNU
-- YACMA autodetected C++ flags: -fdiagnostics-color=auto;-ftemplate-depth=1024;-fvisibility-inlines-hidden;-fvisibility=hidden;-Wno-attributes
-- YACMA autodetected C++ debug flags: -Wall;-Wextra;-Wnon-virtual-dtor;-Wnoexcept;-Wlogical-op;-Wconversion;-Wdeprecated;-Wold-style-cast;-pedantic-errors;-Wdisabled-optimization;-ftemplate-backtrace-limit=0;-Wsuggest-attribute=pure;-Wsuggest-attribute=const;-Wsuggest-attribute=noreturn;-Wsuggest-attribute=format;-Wodr;-Wsuggest-final-types;-Wsuggest-final-methods;-Wsuggest-override;-Wshift-negative-value;-Wshift-overflow=2;-Wduplicated-cond;-Wnull-dereference;-Wno-maybe-uninitialized
-- Thread library: 
-- POSIX threads detected.
-- POSIX threads affinity extensions NOT detected.
-- The 'thread_local' keyword is supported.
-- Python modules do NOT require linking to the Python library.
-- Python interpreter: /usr/bin/python
-- Python interpreter version: 2.7.14
-- Python include dir: /usr/include/python2.7
-- Generic UNIX platform detected.
-- Python modules install path: lib/python2.7/site-packages
-- Required Boost libraries: 
-- Boost version: 1.68.0
-- Detected Boost version: 106800
-- Boost include dirs: /usr/local/include
CMake Error at build/pagmo-config.cmake:16 (include):
  include could not find load file:

    /home/me/pagmo2/build/pagmo_export.cmake
Call Stack (most recent call first):
  CMakeLists.txt:148 (find_package)


-- Required Boost libraries: python27
-- Boost version: 1.68.0
-- Found the following Boost libraries:
--   python27
-- Detected Boost version: 106800
-- Boost include dirs: /usr/local/include
[...]

PR #221, opened 2018-09-08
mp_island: add ability to spawn new processes rather than using a pool

bluescarni:
The ``mp_island`` pythonic island can now, optionally, spawn a new process for each evolution, rather than submitting jobs to a global process pool. This allows to match the behaviour of ``thread_island`` on the C++ side, and might be useful to enforce truly concurrent evolution in archipelagos regardless of the pool size or the number of processors available.

The API has not changed, and the default is for ``mp_island`` to use the pool.

I also started using intersphinx more: now we link correctly to the current documentation for ipyparallel and numpy as well (though I did not change the numpy references in the docs to do the actual hyperlinking, I did it only in one place for testing - it's probably better to update the docs in a separate PR).


bluescarni:
Forgot that this also contains a one-liner to account for the missing linking of the python dll on cygwin, as reported in #216.

PR #226, opened 2018-10-31
(WIP) Batch fitness evaluation

bluescarni:
WIP list of further improvements that need to be tackled after this PR is done (will move this to an issue report eventually):

- implement proper ``batch_fitness()`` support for the ``decompose`` and ``unconstrain`` metaproblems. ``translate`` already supports ``batch_fitness()`` (done in this PR). Not conceptually difficult, but it may require some refactoring of the multiobjective utilities to accommodate the ``batch_fitness()`` data layout. @darioizzo would you like to take care of this eventually?
- the performance of an archipelago's batch init can be improved by generating in batch mode *all* the individuals of the archipelago, rather than one island at a time as we are doing now.
- the python part of the PR is working but incomplete. Specifically:
  - needs more testing (i.e., test the new problem methods, test the ``bfe`` class, test the new island/pop/archi constructors, etc.)
  - the ``bfe`` functionality needs to be exposed in the AP machinery, so that APs can implement their own ``bfe`` in the same way they can implement their own problems/algos/islands
  - there are no pythonic bfes implemented, so the bfe functionality is not available for pythonic problems. We need pythonic bfes based on multiprocessing and ipyparallel, and then we need to override the logic for choosing the "best" bfe type based on the problem properties (in the same way we do for islands)
- we need more documentation for ``bfe``. The API docs are complete, but we need tutorials and user-level docs for the functionality. Perhaps some examples showing how the ``bfe`` machinery improves performance for heavy problems, when cting large pops, etc. An example showing the ``bfe`` working with opencl and/or with vectorised instructions would be great. On the Python side, we could use numba, pyopencl, etc. etc.
  - In user docs heading, we should use something more descriptive than "bfe", "batch evaluators" is probably a good compromise in terms of readability.
- for symmetry with the island class, it would probably make sense to have a ``fork_bfe`` down the line, but not super high priority


bluescarni:
@darioizzo One consequence of this is that now we depend on TBB. This is easy to handle in conda, but we will have to add TBB to the manylinux builds (probably not so difficult) and then the pip mingw windows builds (which is gonna be a mess probably).

darioizzo:
@bluescarni I did already the tbb manylinux build for dcgp you can use that as a baseline. for windows too.

include/pagmo/problem.hpp
```diff
@@ -1962,6 +1962,17 @@ class problem
         return m_fevals.load(std::memory_order_relaxed);
     }
 
+    /// Increment the number of fitness evaluations.
+    /**
+     * This method will increase the internal counter of fitness evaluations by \p n.
+     *
+     * @param n the amount by which the internal counter of fitness evaluations will be increased.
+     */
+    void increment_fevals(unsigned long long n) const
```
darioizzo: where is this to be used?

bluescarni: When implementing batch fitness evaluation, we often have to operate on copies of the input problem (e.g., the problem is not thread safe enough, we are using multiprocessing thus the problem needs to be serialized back and forth, etc.). As a consequence, we will be invoking fitness evaluations on problem instances different from the original one, and the original problem will not "see" the fitness evaluations that happen elsewhere. So what we can do in this situation is to use this helper to manually bump up the number of fitness call performed in the original problem object.

It's not essential, but it's needed if we want to enforce some sort of consistency between ``fitness()`` and ``batch_fitness()``, in terms of measured fevals.

bluescarni:
@darioizzo this is good to go for me, wanna take a look?

include/pagmo/batch_evaluators/member_bfe.hpp
```diff
@@ -0,0 +1,36 @@
...
+for more details.
+
+You should have received copies of the GNU General Public License and the
+GNU Lesser General Public License along with the PaGMO library.  If not,
+see https://www.gnu.org/licenses/. */
+
+#ifndef PAGMO_BATCH_EVALUATORS_MEMBER_BFE_HPP
+#define PAGMO_BATCH_EVALUATORS_MEMBER_BFE_HPP
+
+// NOTE: the member_bfe is defined in bfe.hpp
```
darioizzo: trust you here, but oh god reminds me of the hypervolume circular deps maddness ...

doc/sphinx/docs/cpp/batch_evaluators/default_bfe.rst
```diff
@@ -0,0 +1,53 @@
...
+      :param dvs: the input decision vectors that will be evaluated.
+
+      :return: the fitness vectors corresponding to the input decision vectors in *dvs*.
+
+      :exception std\:\:invalid_argument: if the input problem *p* does not provide a batch fitness member function and it is does not provide at least the :cpp:enumerator:`~pagmo::thread_safety::basic` thread safety level.
+      :exception unspecified: any exception raised by the call operator of :cpp:class:`~pagmo::member_bfe` or :cpp:class:`~pagmo::thread_bfe`.
+
+   .. cpp:function:: std::string get_name() const
+
+      :return: a human-readable name for this :cpp:class:`~pagmo::default_bfe`.
```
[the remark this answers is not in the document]
bluescarni: Because if we ever decide to change the string, we don't have to change the docs too :)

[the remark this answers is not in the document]
bluescarni: Also, no code should ever rely on a specific string being returned by these methods, so by leaving it unspecified we possibly pre-empt wrongly relying on this string staying the same in the future or whatever...

doc/sphinx/docs/python/python_docs.rst
```diff
@@ -16,18 +16,21 @@ These are the core PyGMO classes.
    py_population
    py_island
    py_archipelago
+   py_bfe
    py_misc
 
-Implemented problems, algorithms and islands
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-These are the user-defined problems, algorithms and islands implemented in PyGMO.
+Implemented problems, algorithms, islands and bfes
```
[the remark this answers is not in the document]
bluescarni: Agreed, I was just sick of writing "batch fitness evaluator" over and over :/ "Batch evaluators" works fine, I'll add it.

include/pagmo/batch_evaluators/member_bfe.hpp
bluescarni: Don't worry, this is all gone away in the library branch.

bluescarni:
@darioizzo Thanks for the review! I will address the docs modifications in a separate PR (I have added some more todos in the list above).

Issue #227, opened 2018-11-06
Can I write a thread_safe python user defined problem?


chengdi123000:
I cannot get a problem even I write a `get_thread_safety(self)` function to return `True`.

I want to know if this is a limitation of python API.

chengdi123000:
Thank you for your reply. 

I want it because threads can share a lot of things. Message passing between processes is so painful. 

I have a model fitting problem with about 200 experiment data point. I want to use some optimization method to find the pareto front of two objectives, which are fraction of variance unexplained (FVU)s of two performance index. The model simulation invokes a lot of complex cython objects. So it is difficult to do serialization.

I have two choices: 
1. distribute simulation of every data point to ipyparallel nodes. FVU are calculated in master node
2. distribute FVU calculation to ipyparallel engines. Master node create an island to hold the status.

I prefer the first one, but I am confusing. There are two additional level of parallelism in pagmo2: islands and archipelagos. And pickle is also a source of many errors. I have to define some `__reduce__` method to transfer the user defined problem class.

bluescarni:
At the moment we don't support parallelisation via threads in python.

Note that, as far as CPython is concerned, you can get parallel speedup with multiple threads only if you are manually releasing the GIL whenever you are not using the Python interpreter from the current thread (and of course the computationally intensive parts of your code have to be written in some language other than Python). Are you doing this type of GIL management bits by yourself (or perhaps is Cython is doing it for you)? Because if you are not, you will not get any speedup from multithreaded Python.

You don't have to use ipyparallel necessarily, unless you are on Python 2. If a single computer is enough for your optimisation tasks, then you can use the multiprocessing island (``mp_island``). I believe it is possible to share at least certain data types between processes via shared memory in python, so perhaps if you are having troubles with serialization you could use such facilities to communicate between your multiprocessing islands. See for instance this SO answer:

https://stackoverflow.com/questions/10721915/shared-memory-objects-in-multiprocessing

chengdi123000:
Unfortunately, my program may need more nodes for memory and FLOPs. So ipyparallel is needed. I just implement a `__reduce__` and use `Client.use_pickle` method to transfer the state of problem. It works. The default enhanced pickle library of ipyparallel `cloudpickle` have some trouble with cython objects even when I write a `__reduce__` for user defined problem.

chengdi123000:
But I found that `ipyparallel_island` do not work well with my method because it must use `cloudpickle` to do serialization.

[...]

PR #229, opened 2018-11-09
Allow to select a serialization backend other than cloudpickle

bluescarni:
See also the discussion in #227.


bluescarni:
The dependencies don't change, pygmo still depends hard on cloudpickle (which remains the default backend). If the user has also ``dill`` installed, then she can switch to it as a serialization backend. If the user tries to switch to ``dill`` and ``dill`` is not installed, an exception will be raised.

darioizzo:
ok noted.

PR #230, opened 2018-11-11
Build system fixes/improvements

bluescarni:


bluescarni:
If you are referring to the fact that we were using the Boost version CMake variable before it is actually available, then this is addressed by the current PR: we run ``find_package(Boost)`` twice, the first time we look only for the headers and determine the boost version in the process, the second time we look for the compiled boost libraries we need (with Boost.Python's name depending on the Boost version).

Or were you referring to something else?

darioizzo:
No ok I was referring to that, but I also thought that you wanted to change the logic some how ... nvm.

PR #247, opened 2019-02-12
Lennard Jones and Golomb Problems

darioizzo:
The implementation of two new UDP: the Lennard Jones Cluster problem and the Golomb Ruler.

They both are scalable but scale badly so its useful to have the to test the batch fitness capability in #226


include/pagmo/problems/golomb_ruler.hpp
```diff
@@ -0,0 +1,216 @@
...
+                                                   + std::to_string(order) + " was requested.");
+        }
+        if (upper_bound < 2u) {
+            pagmo_throw(
+                std::invalid_argument,
+                "The upper bound for the maximum distance between consecutive ticks has to be at least 2, while "
+                    + std::to_string(upper_bound) + " was requested.");
+        }
+        // Overflow can occur when evaluating the fitness later if the upper_bound is too large.
+        if (static_cast<double>(upper_bound) * (order - 1u)
```
bluescarni: This overflow check should not involve floating point arithmetics. You can rewrite it as

```c++

if (upper_bound > std::numeric_limits<unsigned>::max() /  (order - 1u)) {
    pagmo_throw(...);
}

```

doc/sphinx/docs/cpp/problems/golomb_ruler.rst
```diff
@@ -0,0 +1,103 @@
...
+   .. cpp:function:: template <typename Archive> void serialize(Archive &ar)
+
+      Object serialization.
+
+      This method will save/load this into the archive **ar**.
+
+      :param ar: target archive.
+      :exception unspecified: any exception thrown by the serialization of the UDP and of primitive types.
+  
+This UDP is typically used to construct a :cpp:class:`~pagmo::problem`.
```
bluescarni: I would just drop this line...

include/pagmo/problems/golomb_ruler.hpp
```diff
@@ -0,0 +1,216 @@
...
+        // 1 - We compute the ticks (the ruler length will be the last tick since we start from 0)
+        vector_double ticks(x.size() + 1, 0.);
+        std::partial_sum(x.begin(), x.end(), ticks.begin() + 1u);
+        f[0] = ticks.back();
+        // 2 - We compute all pairwise distances
+        vector_double distances;
+        for (decltype(ticks.size()) i = 0; i < ticks.size() - 1; ++i) {
+            for (decltype(ticks.size()) j = i + 1; j < ticks.size(); ++j) {
+                auto tmp = ticks[j] - ticks[i];
+                distances.push_back(tmp);
```
bluescarni: Constructing a vector with multiple ``push_back()`` will kill performance, as it will typically result in multiple reallocations to grow the vector a bit at a time. If we care about performance here, we should prepare the ``distances`` vector with the correct size and then index directly into it.

That however raises the issue of possible overflows in the ``size_type`` of ``vector_double``, which we did not check above because we are checking against possible overflows wrt ``unsigned`` instead. So either we need an extra overflow check on construction, or we need to replace the use of ``unsigned`` with ``vector_double::size_type``.

I would also remove the temporary variable ``tmp``, as it is only used once.

include/pagmo/problems/golomb_ruler.hpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+        // 3 - We compute how many duplicate distances are there.
+        std::sort(distances.begin(), distances.end());
```
bluescarni: Here we have the usual issue of potentially sorting with ``NaN`` values. Perhaps we could add a check that the produced distances are all non-``NaN``? Or otherwise we can use one of the custom comparators we have in some utility file (in which case, such a custom comparator should also be used in the ``std::unique()`` call below).

include/pagmo/problems/lennard_jones.hpp
```diff
@@ -0,0 +1,182 @@
...
+The PaGMO library is distributed in the hope that it will be useful, but
+WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+for more details.
+
+You should have received copies of the GNU General Public License and the
+GNU Lesser General Public License along with the PaGMO library.  If not,
+see https://www.gnu.org/licenses/. */
+
+#ifndef PAGMO_PROBLEM_LENNARDJONES_HPP
```
bluescarni: Same consistency issue as above, missing underscore between ``LENNARD`` and ``JONES``.

include/pagmo/problems/lennard_jones.hpp
```diff
@@ -0,0 +1,182 @@
...
+#define PAGMO_PROBLEM_LENNARDJONES_HPP
+
+#include <iostream>
```
bluescarni: We are using ``std::pow()`` below, need to include ``<cmath>``.

include/pagmo/problems/lennard_jones.hpp
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @param x the decision vector.
+     *
+     * @return the fitness of \p x.
+     */
+    vector_double fitness(const vector_double &x) const
+    {
+        vector_double f(1, 0.);
+        double sixth, dist;
+        f[0] = 0;
```
bluescarni: ``f[0] = 0;`` can be dropped, as we already inited ``f`` with a single value of zero above.

include/pagmo/problems/lennard_jones.hpp
```diff
@@ -0,0 +1,182 @@
...
+     * Computes the fitness for this UDP
+     *
+     * @param x the decision vector.
+     *
+     * @return the fitness of \p x.
+     */
+    vector_double fitness(const vector_double &x) const
+    {
+        vector_double f(1, 0.);
+        double sixth, dist;
```
bluescarni: Unless I am mistaken, there is no need to declare these variables outside the for loop. Just create them on the fly within the loop, so it's clear these are not values that need to survive past the loop.

include/pagmo/problems/lennard_jones.hpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+    /// Box-bounds
+    /**
+     * Returns the box-bounds for this UDP.
+     *
+     * @return the lower and upper bounds for each of the decision vector components
+     */
+    std::pair<vector_double, vector_double> get_bounds() const
+    {
+        unsigned prob_dim = 3u * m_atoms - 6u;
```
bluescarni: There should an overflow check here:

```c++
if (m_atoms > std::numeric_limits<unsigned>::max() / 3u) {
    ...
}
```

However there's the same issue here as above, in mixing ``unsigned`` and ``vector_double::size_type``. So either we need 2 overflow checks, or we need to use consistently ``vector_double::size_type`` everywhere.

tests/golomb_ruler.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <pagmo/types.hpp>
+
+using namespace pagmo;
+
+BOOST_AUTO_TEST_CASE(golomb_ruler_test)
+{
+    // Problem construction
+    BOOST_CHECK_THROW((golomb_ruler{0u, 10u}), std::invalid_argument);
+    BOOST_CHECK_THROW((golomb_ruler{10u, 0u}), std::invalid_argument);
+    BOOST_CHECK_THROW((golomb_ruler{2300u, 1421552366u}), std::overflow_error);
```
bluescarni: Would be probably best to use ``std::numeric_limits`` here to trigger the error, in order to make the test portable.

include/pagmo/problems/golomb_ruler.hpp
darioizzo: I will reserve then.

PR #249, opened 2019-02-15
(WIP) gaco

Sceki:
I am Giacomo Acciarini, a 2nd year MSc student at TU Delft. I am currently pursuing an MSc in Aerospace Engineering, track: Space Flight, profile: Space Exploration.

As a part of my thesis I would like to implement an extended ant colony optimization algorithm with oracle penalty method (with possible modifications to be added later). I thus set up a collaboration with Dario Izzo and the ACT for writing the algorithm in such a way that it will eventually be available in an open source way for the entire PaGMO community.


darioizzo:
Ok, so new chapter :). 

While we take a look at the code, you need to start compiling, and from now on keeping it, in Debug mode. This is done in cmake passing or selecting the option ```CMAKE_BUILD_TYPE``` to be ```Debug```.

You will notice that as soon as you do this compilation will fail as your code does a lot of unwanted and dangerous conversions between unsigned, vector_double::size_type, int etc.... 

Start looking into them and fix them and whenever in doubt, please write on our gitter channel (not in here) asking questions or for solutions .... when done commit your fixes in one only commit named "[SKIP CI] - Fixes for debug mode"

Note that the aim is to get no warnings (which we treat as errors) issued during the debug compilation.

In the meantime I will start to have a review of the code (avoiding to highlight all the conversion issues and assuming they will be fixed).

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,995 @@
...
+    typedef std::tuple<unsigned, double, double, double, double, double, double> log_line_type;
+    /// The log
+    typedef std::vector<log_line_type> log_type;
+
+    /**
+     * Constructs the ACO user defined algorithm for single and multi-objective optimization.
+     *
+     *
+     *  @param[in] gen Generations: number of generations to evolve.
+     * @param[in] acc Accuracy parameter: for inequality and equality constraints .
```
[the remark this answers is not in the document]
Sceki: I did not update the description of this parameter. I basically introduced this acc to let the user define the minimum distance between penalty values for a solution to be accepted. I will update the description!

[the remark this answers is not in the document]
darioizzo: Is this not the same meaning as the ```c_tol``` of the pagmo::problem? Why not using it then? We already ask the user to define a tolerance on the constraint satisfaction.

If its not the same, then indeed the documentation must be clearly addressing the semantic difference between the two.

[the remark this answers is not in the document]
Sceki: I was planning to use c_tol for when computing the inequality and equality constraints violations and then to use another tolerance (this acc) for maintaining a certain distance between penalties. 
For instance, if, in mathematical terms, g_e(X)=0, then, in the code, I would have considered g_e(X)=0 also if g_e(X)<=acc. And this would have been done inside the penalty computation. Afterward, I would have also applied a minimum distance between the sorted penalties: so that if a solution is not at least |pen_i - pen_{i-1}|>=acc, then the solution is not accepted.

However, now that you pointed out this, and since these two parameters are somehow related to the same aspects, it is probably more efficient and simpler to directly use c_tol for the penalties, and leave the equality and inequality constraint violations computation in mathematical terms.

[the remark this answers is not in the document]
darioizzo: Did you consider using detail::test_eq_constraints and detail::test_ineq_constraints ? they are in ```pagmo/utils/constrained.hpp``` maybe they can be of use for your case.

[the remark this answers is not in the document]
darioizzo: You can pass to these fnctions c_tol .... and get the L2 violation discounted by c_tol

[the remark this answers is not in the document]
Sceki: Cool! I did not notice them! Thanks!

darioizzo:
Good job again. I will now do a second iteration of code review. What I do is:

1) Check the CI failures and suggest you changes to the code to fix those (from next time you can do this yourself). Except the code coverage, which will be fixed eventually by doing tests properly, its nice if from now on we try to keep travis and appveyor happy . This will ensure that the code is cross platform and does not contain obvious bugs (see for example your tuple definition which makes all the first CI fail ... thats a bug)

2) Check some obvious inefficiencies (such as memory allocations, or unnecessary object instantiations)

In the meantime: we are currently testing on rosenbrock (and we will discuss performances as soon as we cleaned a bit more the code), but the algorithm is currently also able to deal with constraint. Did you check if that part also kind of works? FOr example using ```hock_schittkowsky_71``` or the ```cec2006 suite```?

Sceki:
Thank you for the reviews and suggestions!
- okay for the CI failures: from next time on I will look into them! Do you recommend specific documentation to understand how AppVeyor and Travis CI work?
-I will try to look for those, also referring to other optimization codes in PaGMO.
-I have not checked whether the algorithm works in the presence of constraints. I will do it today using the test suites that you suggested

darioizzo:
No need to learn how they work, just click on the icons and see the virtual machine log (example: https://travis-ci.org/esa/pagmo2/jobs/495191140)

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,974 @@
...
+        stream(ss, "\n\tAccuracy parameter: ", m_acc);
+        stream(ss, "\n\tObjective stopping criterion: ", m_fstop);
+        stream(ss, "\n\tImprovement stopping criterion: ", m_impstop);
+        stream(ss, "\n\tEvaluation stopping criterion: ", m_evalstop);
+        stream(ss, "\n\tFocus parameter: ", m_focus);
+        stream(ss, "\n\tKernel: ", m_ker);
+        stream(ss, "\n\tOracle parameter: ", m_oracle);
+        stream(ss, "\n\tMax number of non-dominated solutions: ", m_paretomax);
+        stream(ss, "\n\tPareto precision: ", m_epsilon);
+        stream(ss, "\n\tDistribution index for generating multi-kernel gaussian pdfs: ", m_e);
```
darioizzo: ```m_e``` is not a distribution index.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,974 @@
...
+        if (m_omega_strategy == 1) {
+            for (decltype(pop_size) j = 0u; j < pop_size; ++j) {
+                vector_double dvs_new_j(
+                    n_con); // here I store all the variables associated with the j_th element of the sol_archive
+
+                for (decltype(n_con) h = 0u; h < n_con; ++h) {
+                    double g_h = 0;
+
+                    for (decltype(sol_archive.size()) k = 0u; k < sol_archive.size(); ++k) {
+                        std::normal_distribution<double> gauss_pdf_1{sol_archive[k][1 + h], sigma[h]};
```
darioizzo: I see here (and later) you are instantiating  several ```std::normal_distribution<double>``` changing the std and mean. I suggest to instantiate only one of these objects (maybe above where you instantiate the uniform_real) ...
```
std::normal_distribution<double> gauss{0., 1.};
```
then you can draw a number anywhere in the code, with generic mu and std as follows:
```
my_normally_distributed_number = mu + std * gauss(m_e);
```
this produces code that is both more elegant and more efficient.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,974 @@
...
+                                  std::setw(15), "dx:", std::setw(15), std::setw(15), "dp:", '\n');
+                        }
+
+                        print(std::setw(7), gen, std::setw(15), fevals, std::setw(15), best_fit, std::setw(15), m_ker,
+                              std::setw(15), worst_fit, std::setw(15), m_oracle, std::setw(15), dx, std::setw(15), dp,
+                              '\n');
+
+                        ++count_screen;
+                        // Logs
+                        m_log.emplace_back(gen, best_fit, m_ker, worst_fit, m_oracle, dx, dp);
```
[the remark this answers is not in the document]
Sceki: Does the prints and log have to correspond? 
If yes, I forgot to add the fevals in the log: I will do it in the next commit. In the meantime, I am testing cec2006 problems.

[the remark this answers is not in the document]
darioizzo: Yes they have to correspond

darioizzo:
@Sceki 
* Do not use M_PI thats not defined in all platform. Use ``` boost::math::constants::pi<double>() ```
* You can run doxygen yourself. Just go to the folder ```doc/doxygen``` and type ```doxygen```

Sceki:
- When I try `boost::math::constants::pi<double>()` it gives me an error (_no member named math in boost_). Can I define it as a normal variable  (i.e., `const double PI  =3.141592653589793238463` )?
- Also, I am currently getting a warning in doxygen because the get_log is not documented. In order to fix this, I would like to know how are the examples generated in the other codes? Is there a specific problem and seed they are referring to, or can it be just a random example

darioizzo:
@Sceki 
* No. You need to use ``` boost::math::constants::pi<double>()```. Include the corresponding header.
* In the docs you can use a random seed to generate the example.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,978 @@
...
+
+        // Here I define the weights. Their definition depends on the user's selection of the m_omega_strategy parameter
+        // (either 1 or 2)
+        if (gen == 1 && m_omega_strategy == 1) {
+            // We declare a vector with 'ker' doubles
+            vector_double ker_vector(m_ker);
+            // We fill it with 1,2,3,...,K
+            std::iota(std::begin(ker_vector), std::end(ker_vector), 1);
+            // We compute the sum of the elements
+            double sum = std::accumulate(ker_vector.begin(), ker_vector.end(), 0);
```
darioizzo: This is a tricky one (and the reason for the good practice of using ```dots``` for double constants ```u``` for unsigned etc...). When you us ```std::accumultate``` and you init with the value ```0```, the compiler deduces its type as integer. He will then start accumulating integers and will return an integer (https://en.cppreference.com/w/cpp/algorithm/accumulate), while ker_vector is full of doubles, so he converts your doubles to integers (https://ci.appveyor.com/project/ci4esa/pagmo2/builds/22527804/job/ls60fetvov7v5m4u#L934). 

This is a plain bug as I guess that is not what you want to do!!

Sceki: Fortunately, I was summing integers even though I defined a vector_double as ker_vector. Thus, the results were not affected by this. I nevertheless fixed this bug by substituting `0` with `0u` as you suggested, and `vector_double ker_vector(m_ker)` with `std::vector <unsigned> ker_vector(m_ker)`.

darioizzo: you may still have issues as sum is ```double``` and some compilers do not like to convert implicitly ints do doubles ... but lets see

Sceki: Maybe defining `sum` as `unsigned` will fix the issue

darioizzo:
YAY! So the CI is now green, do not bother with the test coverage for the time being. I think the next step is to check the code with constrained problems as you did with Rosenbrock. Then integers. Then we take a pause and discuss performances ok?

Sceki:
Hello,

I have run the 24 CEC_2006 test problems. In doing so I followed the rules from the conference (using the same max number of evaluations, etc.).

Overall, I managed to solve:
•  Problems: 1,6,8,12,16 with 100% success and feasibility rate.
•  Problems: 3, 11, 17 with 100% success and 0% feasibility rate.
•  Problems: 4 (far), 7 (near), 9 (near), 10(medium), 18(near), 19(near) with 100% feasibility and 0% success
•  Problems: 5(near), 13(near), 14(near), 15(near), 20(near), 21(far), 22(far), 23(far) with 0% feasibility and 0% success
• Problem 2 with 100% feasibility and 28% success 
• Problem 24 with 100% feasibility and 48% success

When I say near, I mean that the minimum is around <10^-1 wrt the best known solution.

For all these I kept the same parameters for the algorithm, besides the oracle parameter, and I stopped the simulation when the solution was nearer than 10^-4 from the best known solution.
I chose 200 population and 2500 generations, and I would say that the most critical problems might be: 4, 10, 21, 22, 23; whereas the others may turn out to have a high success rate if tweaking and tuning of optimization parameters is applied.

I was wondering how do you want me to handle these problems, shall I push the code I used?
Also, since I am using a solution archive where I store the best individuals throughout the generations, my final solution archive will be the final population, and not the population itself. I was thus wondering whether it could be a good idea to directly return the last archive as last population, considering that this would mean that the last population has size m_ker and not pop_size?

darioizzo:
I am currently away and thus not very responsive. 

The returned pop size must be unchanged. The idea is that you are evolving the chromosome material of the initial population somehow. If during the evolve you manipulate a larger amount of chromosomes, then you need some reinsertion mechanism to put it back into the pop.

Is the algorithm working on integer chromosomes?

Did you check with rastrigin minlp? Or with the golomb ruler?

Sceki:
Hello, I still have to implement the integer part. 
Would it be possible to proceed while I do this in parallel? At this moment, I feel that is also important to make the algorithm work well for continuous variables and I would like to implement the integers at a later phase, if possible. Do you think this could still be an effective workflow?

Maybe I could throw for integers now and come back to them later.

darioizzo:
Giacomo,

Ok, then throw for integers and multiobjective and let's proceed.

1) We have to unit test the code. Have a look to how other tests are done
i.e. sade ... remember that unit tests are to make sure all lines of the
code are run during tests and that they produce the correct result. Since
the evolve method is typically heuristic the unit test for that method are
just made of several calls to the method making sure that the throws are
triggered as well as the different branches.

2) When the unit test coverage is passed, we will expose the new uda to
python

3) We will write a tutorial on python on how to solve the cec2016 with gaco.

Sceki:
Hello.
Sorry, I was away the previous days and could not work on the algorithm. I just pushed the gaco.cpp code where I implemented the unit tests which should work without errors and I fixed small things in the gaco.hpp code.
Do you want to discuss performances after having worked with python?

darioizzo:
Yes, I would like the coverage to be green (you can check what lines are not covered by your test) and then have the python exposition.

In python is much easier to discuss and see perfs ....

Sceki:
Hello, 
AppVeyor seems to fail due to the fact that it takes more than 60 mins to execute certain builds and this makes the check to stop with consequent inconclusive results.

Do you know how could I fix this? I do not really get what could cause such lengthy build times.

bluescarni:
@Sceki The appveyor build sometimes times out, but it is usually a transient error. I will restart the build now.

Longer term, we have some plans on how to reduce the time needed for each build job.

darioizzo:
@Sceki 
also, I noticed your test last 400 seconds ... for udas tests should take approximately 1 sec .... there is no need to let the volve run for many generations on large populations .... test on a 2-3 generations and small pops ....

bluescarni:
I just restarted one of the builds in travis which failed due to a network timeout.

darioizzo:
@Sceki Congrats!! Its all green Level 2 is passed :)
From now on you should aim, at each commit to preserve this all green status, so check CI failiures and fix them or ask us for advise.

Next step should be relatively short and simple:

* Expose gaco uda to python. This is done adding code to the file ```/pygmo/expose_algorithms_1.cpp```, see ```https://github.com/esa/pagmo2/blob/master/pygmo/expose_algorithms_1.cpp#L89``` as an example. In general a stochastic uda such as gaco has two constructors exposed (allowing to set and control the seed), and the log. Any other public method that you think may be needed by users need to be exposed explicitly. Note also that you have to write the docstring (see for inspiration those of other udas, and be careful to respect the REST syntax we use)

Sceki:
Thanks for the suggestions. I will start exposing my algorithm to PyGMO!

It is probably a silly question, but how can I build and run the expose_algorithm_1.cpp file from the terminal? Should I do a similar procedure as I did for gaco (e.g., `ccmake ../` etc.) but now cding inside the pygmo directory?

darioizzo:
So you need to

1. Install the pagmo headers typing make install

2. In cmake deselect build pagmo and select build pygmo

3. Make build  will now build pygmo (the python binding) and make install
will install the pygmo module in python

Make sure the cmake root is your .local so that the installation will be
local to your user

Sceki:
Hello, 
when I try to make install the `expose_algorithms_1.cpp` file (where I added my gaco) outputs a very long build error output. One of them is:
```
/usr/local/include/boost/python/signature.hpp:218:1: note: candidate template ignored: could not match 'RT
      (ClassT::*)(T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14)' against 'void (*)(_object *,
      unsigned int, unsigned int, double, double, double, unsigned int, double, unsigned int, unsigned int, unsigned
      int, unsigned int, double, unsigned int, double, unsigned int)'
get_signature(RT(BOOST_PYTHON_FN_CC ClassT::*)(BOOST_PP_ENUM_PARAMS_Z(1, N, T)) Q)
```
and 
```
/usr/local/include/boost/python/signature.hpp:236:1: note: candidate function template not viable: requires 2
      arguments, but 1 was provided
get_signature(
```
I thus tried to remove one of the 15 parameters in my `gaco_.def` definition and it seems to work. SO , if for instance I do make install with:
```
    gaco_.def(bp::init<unsigned, unsigned, double, double, double, unsigned, double, unsigned, unsigned, unsigned,
                       unsigned, double, unsigned, double, unsigned>(
        (bp::arg("gen") = 100u, bp::arg("ker") = 63u, bp::arg("q") = 1.0, bp::arg("oracle") = 0., bp::arg("acc") = 0.01,
         bp::arg("threshold") = 1u, bp::arg("fstop") = 0.000000000001, bp::arg("n_gen_mark") = 7u,
         bp::arg("omega_strategy") = 2u, bp::arg("impstop") = 100000u, bp::arg("evalstop") = 100000u,
         bp::arg("focus") = 0., bp::arg("paretomax") = 10u, bp::arg("epsilon") = 0.9, bp::arg("seed"))));
```
then it outputs those errors, whereas if I do:
```
    gaco_.def(bp::init<unsigned, unsigned, double, double, double, unsigned, double, unsigned, unsigned, unsigned,
                       unsigned, double, double, unsigned>(
        (bp::arg("gen") = 100u, bp::arg("ker") = 63u, bp::arg("q") = 1.0, bp::arg("oracle") = 0., bp::arg("acc") = 0.01,
         bp::arg("threshold") = 1u, bp::arg("fstop") = 0.000000000001, bp::arg("n_gen_mark") = 7u,
         bp::arg("omega_strategy") = 2u, bp::arg("impstop") = 100000u, bp::arg("evalstop") = 100000u,
         bp::arg("focus") = 0., bp::arg("epsilon") = 0.9, bp::arg("seed"))));
```
(where I removed the `paretomax` parameter) it doesn't. I thus think that the problem is that 15 inputs are too many. Do you know a way how I could fix this, or am I forced to remove one parameter from the constructor?

darioizzo:
See: https://stackoverflow.com/questions/8300058/boost-python-limits-to-the-number-of-arguments-when-wrapping-a-function

The limit is 15 but can be risen

In general is a good idea to limit the number of parameters, I was about to comment that if you can remove some its better. Otherwise just raise the limit. We will later see how to simplify the API to modify the algo parameters.

darioizzo:
@sceki Now, install pygmo in your machine and import it in a python session. Try something like:
```
from pygmo import test
test.run_test_suite()
```

Sceki:
I ran the lines you suggested and I got all `ok` as outputs.
Namely, I got the following output:

```
>>> from pygmo import test
>>> test.run_test_suite()
runTest (pygmo.test.core_test_case) ... ok
runTest (pygmo._problem_test.problem_test_case) ... ok
runTest (pygmo._algorithm_test.algorithm_test_case) ... ok
runTest (pygmo._island_test.island_test_case) ... ok
runTest (pygmo._island_test.mp_island_test_case) ... ok
runTest (pygmo._island_test.ipyparallel_island_test_case) ... Waiting for connection file: ~/.ipython/profile_default/security/ipcontroller-client.json
ok
runTest (pygmo.test.de_test_case) ... ok
runTest (pygmo.test.pso_test_case) ... ok
runTest (pygmo.test.pso_gen_test_case) ... ok
runTest (pygmo.test.bee_colony_test_case) ... ok
runTest (pygmo.test.compass_search_test_case) ... ok
runTest (pygmo.test.sa_test_case) ... ok
runTest (pygmo.test.moead_test_case) ... ok
runTest (pygmo.test.sga_test_case) ... ok
runTest (pygmo.test.ihs_test_case) ... ok
runTest (pygmo.test.population_test_case) ... ok
runTest (pygmo.test.archipelago_test_case) ... ok
runTest (pygmo.test.null_problem_test_case) ... ok
runTest (pygmo.test.hypervolume_test_case) ... ok
runTest (pygmo.test.mo_utils_test_case) ... ok
runTest (pygmo.test.con_utils_test_case) ... ok
runTest (pygmo.test.global_rng_test_case) ... ok
runTest (pygmo.test.estimate_sparsity_test_case) ... ok
runTest (pygmo.test.estimate_gradient_test_case) ... ok
runTest (pygmo.test.random_decision_vector_test_case) ... ok
runTest (pygmo.test.cmaes_test_case) ... ok
runTest (pygmo.test.xnes_test_case) ... ok
runTest (pygmo.test.dtlz_test_case) ... ok
runTest (pygmo.test.cec2006_test_case) ... ok
runTest (pygmo.test.cec2009_test_case) ... ok
runTest (pygmo.test.cec2013_test_case) ... ok
runTest (pygmo.test.cec2014_test_case) ... ok
runTest (pygmo.test.luksan_vlcek1_test_case) ... ok
runTest (pygmo.test.minlp_rastrigin_test_case) ... ok
runTest (pygmo.test.translate_test_case) ... ok
runTest (pygmo.test.decompose_test_case) ... ok
runTest (pygmo.test.unconstrain_test_case) ... ok
runTest (pygmo.test.mbh_test_case) ... ok
runTest (pygmo.test.cstrs_self_adaptive_test_case) ... ok
runTest (pygmo.test.decorator_problem_test_case) ... ok
runTest (pygmo.test.nlopt_test_case) ... 
 objevals:        objval:      violated:    viol. norm:
         1         127879             18        674.362 i
         6        13462.5             18        71.0216 i
        11        578.251             18        14.0787 i
        16      0.0674133              3    0.000804609 i
        21    1.27542e-21              0              0

Optimisation return status: NLOPT_XTOL_REACHED (value = 4, Optimization stopped because xtol_rel or xtol_abs was reached)

 objevals:        objval:      violated:    viol. norm:
         1         106178             18        496.627 i
         6        4664.15             18        38.0429 i
        11        43.4688              3     0.00407209 i
        16       0.035541              2     0.00778625 i
        21    3.73081e-19              0              0

Optimisation return status: NLOPT_XTOL_REACHED (value = 4, Optimization stopped because xtol_rel or xtol_abs was reached)
ok
runTest (pygmo.test.ipopt_test_case) ... 
******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt
******************************************************************************


 objevals:        objval:      violated:    viol. norm:
         1         185786             18        759.442 i
         6        3505.85             18        22.2002 i
        11        12.3846             12       0.404696 i
        16    3.76322e-10              0              0

Optimisation return status: Solve_Succeeded (value = 0)

 objevals:        objval:      violated:    viol. norm:
         1        65482.4             18        464.803 i
         6         112.17             18        4.17752 i
        11        56130.3              6       0.400708 i
        16        6.53822              4      0.0257117 i
        21        6.23246              0              0

Optimisation return status: Solve_Succeeded (value = 0)
ok

----------------------------------------------------------------------
Ran 42 tests in 27.987s

OK
```

However, at the first try it was giving errors when I was typing `from pygmo import test`, and I thus tried to follow again the steps you mentioned two comments above, but I got again the errors that I was discussing with Francesco in ghitter. Namely, when I try to configure PyGMO I get the following:
```
CMake Error at
 /Applications/CMake.app/Contents/share/cmake-3.10/Modules/FindBoost.cmake:1928
  (message):
   Unable to find the requested Boost libraries.

   Boost version: 1.68.0

   Boost include path: /usr/local/include

   Could not find the following Boost libraries:

           boost_python36

   No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
   directory containing Boost libraries or BOOST_ROOT to the location of Boost.
 Call Stack (most recent call first):
   cmake_modules/PagmoFindBoost.cmake:22 (find_package)
   CMakeLists.txt:158 (include)



 CMake Error at cmake_modules/PagmoFindBoost.cmake:24 (message):
   Not all requested Boost components were found, exiting.
 Call Stack (most recent call first):
   CMakeLists.txt:158 (include)
```
I do not know why these errors have popped up again, since yesterday I finally managed to configure and generate after I modified the `CMakeCache.txt` file.

darioizzo:
pagmo and pygmo need to be easily installable and testable if you want to be able to develop efficiently.

In a healthy system all you need to do is configure and generate. All libraries and path are found automatically and correctly and make install works out of the box. Our virtual machines (also osx) are all building using and testing pagmo and pygmo almost everyday. Your issues indicate you are in need to clean up your environment and toolchains.

Let me know if you figure it out.

Sceki:
Okay, now everything should be fine.  We can proceed.

darioizzo:
Ok, next step Is to build your docs locally. Install doxygen sphinx and other stuff, go in the doc directory, sphinx and run the script compile.

```pip install 'sphinx<1.7' breathe requests[security] sphinx-bootstrap-theme```

When successfully the docs will be built in the directory _build ... open Index.html using a browser, inspect and report success

**Comment by Sceki** [comment]
@darioizzo 
I followed the steps and opened the index.html file which shows the "Welcome to gaco_project's documentation" with indices and tables. I guess everything went fine!

**Comment by darioizzo** [comment]
Well the idea is to build pagmo documentation you should see the website here: https://esa.github.io/pagmo2/index.html

Let me have a look to the current PR status.

**Comment by Sceki** [comment]
Is this done by modifying/creating .rst.txt files that I have under the `_sources` directory?

**Comment by darioizzo** [comment]
absolutely not. those files are generated by sphinx and should not be touched.

File: include/pagmo/algorithms/gaco.hpp [hunk 6]
```diff
@@ -0,0 +1,899 @@
...
+ * based on three parameters (i.e., pheromone values) which are computed depending on the quality
+ * of each previous solution. The solutions are ranked through an oracle penalty method.
+ *
+ *
+ * The version implemented in pagmo can be applied to box-bounded multiple-objective optimization.
+ *
+ * See:  M. Schlueter, et al. (2009). Extended ant colony optimization for non-convex
+ * mixed integer non-linear programming. Computers & Operations Research.
+ */
+class g_aco
```
darioizzo: class name must be gaco as the file. correct everywhere.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,899 @@
...
+public:
+    /// Single entry of the log (gen, fevals, best_pen, best_fit, m_ker, worst_fit, m_oracle, dx, dp)
+    typedef std::tuple<unsigned, unsigned, double, double, unsigned, double, double, double, double> log_line_type;
+    /// The log
+    typedef std::vector<log_line_type> log_type;
+
+    /**
+     * Constructs the ACO user defined algorithm for single and multi-objective optimization.
+     *
+     * @param[in] gen Generations: number of generations to evolve.
```
darioizzo: the order of these params must be the same as the order in the constructror args.

pygmo/docstrings.cpp
```diff
@@ -2336,6 +2336,86 @@ See also the docs of the relevant C++ method :cpp:func:`pagmo::nsga2::get_log`.
 )";
 }
 
+std::string gaco_docstring()
+{
+    return R"(__init__(gen = 1, ker = 63u, q = 1.0, oracle = 0., acc = 0.01, threshold = 1u,
+fstop = 0.000000000001, n_gen_mark = 7u, impstop = 100000u, evalstop = 100000u, focus = 0.,
+paretomax = 10u, epsilon = 0.9, seed = random)
+
+Giacomo's Ant Colony Optimization algorithm (GACO).
```
darioizzo: Use the same text as in the c++ doxygen strings. No reference to Giacomo, rather to the paper from Schueler.

pygmo/expose_algorithms_1.cpp
```diff
@@ -189,6 +190,22 @@ void expose_algorithms_1()
                nsga2_get_log_docstring().c_str());
 
     nsga2_.def("get_seed", &nsga2::get_seed, generic_uda_get_seed_docstring().c_str());
+    // GACO
+    auto gaco_ = expose_algorithm_pygmo<g_aco>("g_aco", gaco_docstring().c_str());
```
darioizzo: name has to be gaco

darioizzo:
@Sceki I have downloaded and compiled succesfully the PR. I have generated the docs and indeed pagmo documentation was built. I am not sure what you are doing, but it probably boils down again to your setup .... anyway here is more precisely what you have to do

1) intsall the necessary packages (sphinx breathe doxygen, etc...) as detailed above
2) go in the directory ```pagmo2/doc/sphinx``` where pagmo2 is the ROOT of this git repo
3) type ```./compile```
4) while you wait, you should see 
```
(dcgp) Darios MacBook Pro:sphinx darioizzo$ ./compile.sh 
rm -rf _build/*
```
5) go in the directory ```pagmo2/doc/sphinx/_build/html``` and in a browser visualize the file ```index.html```. On my mac the URL is ```file:///Users/darioizzo/Documents/pagmo2/doc/sphinx/_build/html/index.html```

In yours should read something like ```file:///Users/giacomo/Documents/pagmo2/doc/sphinx/_build/html/index.html```

Let me know

Sceki:
Okay. I found the `conf.py` file, but now I get this error when I compile:
```
Configuration error:
There is a syntax error in your configuration file: bad input (conf.py, line 31)
Did you change the syntax from 2.x to 3.x?
```

darioizzo:
The ```conf.py``` file is generated by cmake using the file ```conf.py.in``` included in this git repo. It gets generated when you generate the files in cmake. 

If I were you I would start from zero, you may be accumulating too many weird actions. Delete everything.  Checkout the git repo again. Generate the files in cmake and call compile. I juist did it on my mac and works like a charm.

Sceki:
Would it be possible to clone and make pagmo again in another folder and then transfer the branch gaco there? Something like `\Users\giacomoacciarini\pagmo2` or will this mess up too much with the branches and with the PR?

darioizzo:
From your question it seems to me you have not understood too well what git is and does. Maybe you should follow some tutorials online on the git usage.

Going back to our issue. Follow the following steps literally ... do not skip any:

1) Delete all pagmo2 folders in your machine. use ```rf -fr *```. ALL OF THEM.
2) Go to ```\Users\giacomoacciarini\Documents```
3) type ```git clone https://github.com/Sceki/pagmo2.git``` 
4) ```cd pagmo2``` (THIS WILL BE THE ONLY PAGMO2 DIRECTORY IN YOU SYSTEM)
5) ```git remote add upstream https://github.com/esa/pagmo2.git``` (HERE YOU SET THE ESA GIT REPO AS UPSTREAM FOR POSSIBLE FUTURE MERGES)
6) ```git checkout --track origin/gaco``` (HERE YOU CHECKOUT THE GACO BRANCH OF YOUR GIT REPO)
7) ```mkdir build``` (THIS IS THE ONLY BUILD DIRECTORY YOU WILL NEED)
8) cd build

Now build and install pagmo (not pygmo, only pagmo). In doing so you only have to choose the appropriate ```CMAKE_INSTALL_PREFIX```. Nothing else. You should not EVER EVER touch manually any file nor any configuration .... anything at all (except of course the hpp and cpp files when you develop).

If you use a conda environment activate it before configuring and set ```CMAKE_INSTALL_PREFIX``` to your conda env root (example ```/Users/giacomo/miniconda3/envs/pagmo```)

If you use Homebrew make sure boost and other deps are installed and found.

Let me know,

Sceki:
Okay, the index.html now looks like the website. 
Thanks a lot for all your time and help!

darioizzo:
Ok great, did you follow the steps above? So that I can assume you have exactly that setup in your machine? Also are you using conda or homebrew?

Sceki:
Yes, I followed all the steps, and I used conda environment (`/Users/giacomoacciarini/anaconda3/envs/pagmo`)

Sceki:
I updated CMake to 3.14 but I still get the same error:

```
 CMake Error at
 /usr/local/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:137
 (message):
   Could NOT find Eigen3 (missing: EIGEN3_INCLUDE_DIR EIGEN3_VERSION_OK)
   (Required is at least version "2.91.0")
 Call Stack (most recent call first):
   /usr/local/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:378
 (_FPHSA_FAILURE_MESSAGE)
   cmake_modules/FindEigen3.cmake:92 (find_package_handle_standard_args)
   /Users/giacomoacciarini/anaconda3/lib/cmake/pagmo/pagmo-config.cmake:9
 (find_package)
   CMakeLists.txt:148 (find_package)
```

darioizzo:
After you have built pagmo and installed the headers you need to:
1) in the same build directory as before ```pagmo2/build``` type ```ccmake ../```.
2) deselect the build pagmo and select build pygmo
3) configure and generate.

type ```make -j5```.

Sceki:
If I do it, I still get this error when configuring:
```
 CMake Error at
 /usr/local/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:137
 (message):
   Could NOT find Eigen3 (missing: EIGEN3_INCLUDE_DIR EIGEN3_VERSION_OK)
   (Required is at least version "2.91.0")
 Call Stack (most recent call first):
   /usr/local/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:378
 (_FPHSA_FAILURE_MESSAGE)
   cmake_modules/FindEigen3.cmake:92 (find_package_handle_standard_args)
   /Users/giacomoacciarini/anaconda3/lib/cmake/pagmo/pagmo-config.cmake:9
 (find_package)
   CMakeLists.txt:148 (find_package)
``
Which  `CMAKE_INSTALL_PREFIX`  should I specify? Right now I was using `/Users/giacomoacciarini/anaconda3/envs/pagmo`, the same I used for pagmo

**Comment by darioizzo** [comment]
Write in gitter not here.

**Comment by Sceki** [comment]
Ok I fixed it: I manually specified the path for Eigen3, IPOPT and NLOPT both include files and now configure and generate work. However, when I then do `majke -j5`, I get the following error:
```
Scanning dependencies of target core
[ 12%] Building CXX object pygmo/CMakeFiles/core.dir/expose_algorithms_0.cpp.o
[ 25%] Building CXX object pygmo/CMakeFiles/core.dir/core.cpp.o
[ 37%] Building CXX object pygmo/CMakeFiles/core.dir/docstrings.cpp.o
[ 50%] Building CXX object pygmo/CMakeFiles/core.dir/expose_algorithms_1.cpp.o
[ 62%] Building CXX object pygmo/CMakeFiles/core.dir/expose_problems_0.cpp.o
[ 75%] Building CXX object pygmo/CMakeFiles/core.dir/expose_problems_1.cpp.o
/Users/giacomoacciarini/Documents/pagmo2/pygmo/expose_algorithms_0.cpp:78:10: fatal error: 
      'IpTNLP.hpp' file not found
#include <IpTNLP.hpp>
         ^~~~~~~~~~~~
1 error generated.
```
I am trying to figure out why now.

**Comment by Sceki** [comment]
Due to Travis ci failure, I ran `conda install -c conda-forge boost-cpp` in my pagmo environment: in the next commit I will verify whether this fixes the build failure.

**Comment by darioizzo** [comment]
The CI failiure is not related to your library setup (nor it could be, those are virtual machines that know nothing about your particular setup). If you read the log on travis it clearly suggested to restart the job as some internet problem occurred. I just did that.

Now, your command ```conda install -c conda-forge boost-cpp``` installs boost libraries in your conda pagmo environment. Which suggests me that you were using homebrew libraries before (i.e. in ```\usr\local```)? Please, please do not mix homebrew libraries with conda libraries, it is only going to introduce more problems.

**Comment by Sceki** [comment]
Indeed, as you said. I did not have experience with this and I thus took a while to figure it out, but I think that the problem was caused by the CMake, basically I did not installed it in my conda environment and the program was using that of `\usr\local`. Hence causing still 'bridges' between environment. Thanks again, we can move forward w/o issues of this kind now (hopefully).

**Comment by darioizzo** [comment]
Just to be clear: none of the problems you reported were caused by the use of homebrew cmake. But it is good practice to use everything from conda.

**Comment by darioizzo** [comment]
Good progress!! After fixing CI and coverage, the next step is to allow an integer part for the chromosome, remove all refs to multi objective problems write a tutorial. Then I would merge this into master and start working on the UDP. Do you agree?

**Comment by Sceki** [comment]
Thanks! Yes I agree, just a few things:

• for CI and Appveyor: I saw that appveyor throws an error due to the division by zero, which I used for generating a -inf in the udp of the tests. What do you usually use for generating inf in these cases (w/o failing appveyor)?

• I am not sure about the integer part. I was thinking perhaps to start from MO and then eventually move to integers. But if you think this is a better workflow I trust you

• For the tutorials, do you mean something to be put here: http://esa.github.io/pygmo/tutorials/index.html ? I do not really know how to do it (for how to structure the contents I can look up to the examples there). Is it done through sphinx?

**Comment by darioizzo** [comment]
* For inf use ```std::numeric_limits::infinity```, for nans ```std::numeric_limits::quiet_NaN```
* Multiobjective has a number of complications that is better to face later (like the various utils and the definition of pareto dominance, crowding, eps comparison hypervolume etc.......). So I suggest to start with ithe integers where we all the implementations pertain to ```gaco.hpp```
* You pointed to the old docs. Tutorial are here: ```https://esa.github.io/pagmo2/docs/python/tutorials/python_tut.html``` they are made using sphinx (you can have a look at this file for inspiration: https://github.com/esa/pagmo2/blob/master/doc/sphinx/docs/python/tutorials/cec2013_comp.rst)

I suggest a tutorial on gaco, including some of the plots you have produced. The tutorial should not be in the text advocating that gaco is better than something else, it should just be presenting the API and its use walking the user through code snippets that build the plots and benchmarks ...

File: tests/gaco.cpp [hunk 10]
```diff
@@ -240,10 +241,34 @@ struct udp_inf {
...
+        return {{0., 0.}, {1., 1.}};
+    }
+    /// Problem dimensions
+    unsigned int m_dim;
+};
+
 BOOST_AUTO_TEST_CASE(test_for_inf)
 {
     problem prob(udp_inf);
+    problem prob_2(udp_nan);
```
Sceki: I just realized that these two lines are completely useless. I will erase them in the next commit

Sceki:
@bluescarni , @darioizzo 
Hello, I do not well understand why Travis CI is failing. I can see that a similar failure is occurring in the master branch, so I guess it is not related to me, right?

bluescarni:
@Sceki it's a problem with our documentation toolchain, which is based on a stack of tools whose interplay can be at times brittle.

In this specific case, for a variety of reasons we are forced to use a particular version of sphinx (1.6) which is now getting too old, and it is apparently incompatible with recent versions of breathe (>= 4.12).

I have been pushing lately to abandon the current Doxygen -> Breathe -> Sphinx toolchain in favour of writing the documentation directly in Sphinx, but I haven't managed to convince @darioizzo so far.

You can probably ignore this error for the time being, we'll eventually fix it in some way.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,978 @@
...
+ * ACO is inspired by the natural mechanism with which real ant colonies forage food.
+ * This algorithm has shown promising results in many trajectory optimization problems.
+ * The first appearance of the algorithm happened in Dr. Marco Dorigo's thesis, in 1992.
+ * ACO generates future generations of ants by using the a multi-kernel gaussian distribution
+ * based on three parameters (i.e., pheromone values) which are computed depending on the quality
+ * of each previous solution. The solutions are ranked through an oracle penalty method.
+ *
+ *
+ * The version implemented in pagmo can be applied to box-bounded multiple-objective optimization and its implementation
+ * is an extension of Schlueter's originally proposed ACO algorithm.
```
Sceki: @darioizzo 
I am currently working in the documentation. I thought it could be nice to refer also to my MSc thesis, so that maybe who will be interested can also check it out. 
I was thus thinking to substitute line 60 with a note, in which I would write something like:

The ACO version implemented in PaGMO is an extension of Schlueter's originally proposed ACO algorithm and it was developed by Giacomo Acciarini in collaboration with the PaGMO development team, during his Aerospace Engineering MSc Thesis in TU Delft.

Would this be okay/appropriate?

Sceki:
@bluescarni , @darioizzo 
Hello, yesterday I committed the updated codes together with docs and tutorials.. Travis CI failed (I think not due to my code but for some timeouts).. could you restart it? Also, could you manually make the coverage test trigger, since they did not activate?

bluescarni:
The coverage is not computed if the builds are not fully green, and the documentation build is failing because of problems with sphinx/breathe. @darioizzo is working on a PR for that, when it's merged you will have to merge master into the PR and then the coverage should work.

Sceki:
@bluescarni 
I see, thanks. I thought that the coverage could be triggered anyway since in the commit: "Revert "files to commit before update"", it was triggered even though Travis CI failed

bluescarni:
@Sceki the doc build issues should be solved now, please update the master branch and merge it into your PR.

include/pagmo/algorithms/gaco.hpp
darioizzo: No, when the thesis is online you may do a PR adding the link to it, but for the time being it only creates a broken link (the thesis is not available).

doc/sphinx/docs/algorithm_list.rst
```diff
@@ -27,6 +27,7 @@ Heuristic Global Optimization
 Common Name                                                Docs of the C++ class                     Docs of the python class                  Capabilities
 ========================================================== ========================================= ========================================= =========================
 The null algorithm [#null_algo]_                           :cpp:class:`pagmo::null_algorithm`        :class:`pygmo.null_algorithm`             SM-CU-I-sto
+Ant Colony Optimization (ACO)                              :cpp:class:`pagmo::gaco`                  :class:`pygmo.gaco`                       S-CU-I
```
darioizzo: ACO is reserved for other algorithms that are not this extended version. Use (extended ACO)

doc/sphinx/docs/python/tutorials/cec2006_gaco_benchmark.rst
```diff
@@ -0,0 +1,143 @@
+.. _py_tutorial_cec2006_gaco_benchmark_comp:
+
+ACO Benchmark Using the CEC2006 Problems
```
darioizzo: Again use extended ACO or gaco .... this algorithm is not what the scientific community would recognise as ACO

pygmo/docstrings.cpp
```diff
@@ -2373,6 +2373,104 @@ See also the docs of the relevant C++ method :cpp:func:`pagmo::nsga2::get_log`.
...
+
+ACO is inspired by the natural mechanism with which real ant colonies forage food.
+This algorithm has shown promising results in many trajectory optimization problems.
+The first appearance of the algorithm happened in Dr. Marco Dorigo's thesis, in 1992.
+ACO generates future generations of ants by using the a multi-kernel gaussian distribution
+based on three parameters (i.e., pheromone values) which are computed depending on the quality
+of each previous solution. The solutions are ranked through an oracle penalty method.
+
+
+The version implemented in pagmo can be applied to box-bounded multiple-objective optimization and its implementation is an extension of Schlueter's originally proposed ACO algorithm.
```
darioizzo: Schlueter algorithm is called extended ACO. you may want to summarize briefly here why is our implementation differing from his.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+        auto n_ec = prob.get_nec();
+        auto n_ic = prob.get_nic();
+        auto n_f = prob.get_nf(); // n_f=prob.get_nobj()+prob.get_nec()+prob.get_nic()
+
+        // Other useful variables are stored:
+        std::vector<vector_double> sol_archive(m_ker, vector_double(1 + n_x + n_f, 1));
+        vector_double omega(m_ker);
+        vector_double prob_cumulative(m_ker);
+        std::uniform_real_distribution<> dist(0, 1);
+        std::normal_distribution<double> gauss{0., 1.};
```
bluescarni: For consistency with the previous line, I would not write ``std::normal_distribution<double>`` but just ``std::normal_distribution<>``.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            // inequality constraints are concatenated,for each individual
+
+            // I check whether the maximum number of function evaluations or improvements has been exceeded:
+            if (m_impstop != 0 && m_n_impstop >= m_impstop) {
+                std::cout << "max number of impstop exceeded" << std::endl;
+                return pop;
+            }
+
+            if (m_evalstop != 0 && m_n_evalstop >= m_evalstop) {
+                std::cout << "max number of evalstop exceeded" << std::endl;
```
bluescarni: I would eliminate this line. If I understand correctly, this is just a stopping criterion, no need to be chatty about it.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+                return pop;
+            }
+
+            // 1 - compute penalty functions
+
+            // I declare some useful variables:
+            vector_double penalties(pop_size);
```
bluescarni: Can you declare this variable outside the main loop, pre-allocating the necessary space? So that you allocate the vector once, instead of allocating it from scratch at every loop iteration.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+
+            for (decltype(pop_size) i = 0u; i < pop_size; ++i) {
+                // Penalty computation is here executed:
+                penalties[i] = penalty_computation(fit[i], pop, n_obj, n_ec, n_ic);
+            }
+
+            // 2 - update and sort solutions in the sol_archive, based on the computed penalties
+
+            // I declare a vector where the penalties are sorted:
+            vector_double sorted_penalties(penalties);
```
bluescarni: Same here, can you declare this vector outside the loop? Or perhaps can you just use a single vector, fill it in and sort it at every loop iteration?

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            // This sorts the penalties from the smallest ([0]) to the biggest ([end])
+            std::sort(sorted_penalties.begin(), sorted_penalties.end(), detail::less_than_f<double>);
+            // I declare a vector where I will store the positions of the various individuals:
```
bluescarni: Same here.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            std::vector<decltype(penalties.size())> sort_list(penalties.size());
+            // We fill it with 0,1,2,3,...,K
+            std::iota(std::begin(sort_list), std::end(sort_list), 0);
```
bluescarni: This usage of ``iota`` is incorrect. You need to use not ``0``, but ``decltype(penalties.size())(0)``, because otherwise ``iota`` uses signed integers for the generation of the sequence.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+
+                    ++count_screen;
+                    // Logs
+                    m_log.emplace_back(gen, m_fevals, best_fit, m_ker, m_oracle, dx, dp);
+                }
+            }
+
+            // 4 - compute pheromone values
+
+            vector_double sigma(n_x);
```
bluescarni: Same here.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+        vector_double omega(m_ker);
+        vector_double prob_cumulative(m_ker);
+        std::uniform_real_distribution<> dist(0, 1);
+        std::normal_distribution<double> gauss{0., 1.};
+
+        // PREAMBLE-------------------------------------------------------------------------------------------------
+        // We start by checking that the problem is suitable for this
+        // particular algorithm.
+
+        if (!pop_size) {
```
bluescarni: For an empty population, there's no need to throw. Just return the original (empty) population immediately.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            }
+
+            // 4 - compute pheromone values
+
+            vector_double sigma(n_x);
+            pheromone_computation(gen, prob_cumulative, omega, sigma, pop, sol_archive);
+
+            // 5 - use pheromone values to generate new ants (i.e., individuals)
+
+            // I create the vector of vectors where I will store all the new ants which will be generated:
```
bluescarni: Same.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            std::vector<vector_double> new_ants(pop_size, vector_double(n_x, 1));
+            generate_new_ants(popold, dist, gauss, prob_cumulative, sigma, new_ants, sol_archive);
+
+            for (population::size_type i = 0; i < pop_size; ++i) {
+                vector_double ant(n_x);
```
bluescarni: Same here. If possible and if it does not hurt the readability of the code, it would be better to have some temp vectors declared outside of the loop and re-used in these inner loops.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+            // I reset the impstop counter since the solution archive is updated:
+            m_n_impstop = 1;
+            // I save the variables, objectives, eq. constraints, ineq. constraints and penalties in three different
+            // vectors:
+            for (decltype(m_ker) i = 0u; i < m_ker; ++i) {
+                variables[i] = pop.get_x()[sorted_list[i]];
+                fitness[i] = pop.get_f()[sorted_list[i]];
+                temporary_penalty[i] = sol_archive[i][0];
+            }
+            std::vector<unsigned> saved_value_position;
```
bluescarni: Same general comments as above, would be good to avoid re-creating vectors at every loop iteration.

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+
+        // No throws, all valid: we clear the logs
+        m_log.clear();
+
+        // Main ACO loop over generations:
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) {
+            // At each generation we make a copy of the population into popold
+            population popold(pop);
+
+            auto dvs = pop.get_x(); // note that pop.get_x()[n][k] goes through the different individuals of the
```
bluescarni: If I am reading the code correctly, ``dvs`` and ``fit`` are not written to inside the loop. If that's the case, then you can mark them ``const`` and create references to them, so that you avoid copying:

```c++
const auto &dvs = pop.get_x();
const auto &fit = pop.get_f();
```

include/pagmo/algorithms/gaco.hpp
```diff
@@ -0,0 +1,988 @@
...
+     * This method will save/load \p this into the archive \p ar.
+     *
+     * @param ar target archive.
+     *
+     * @throws unspecified any exception thrown by the serialization of the UDP and of primitive types.
+     */
+    template <typename Archive>
+    void serialize(Archive &ar)
+    {
+        ar(m_gen, m_acc, m_impstop, m_evalstop, m_focus, m_ker, m_oracle, m_paretomax, m_epsilon, m_e, m_seed,
```
bluescarni: I count 24 data members in this class, but it seems like only 23 of them are serialized. Did you forget some members in this call?

bluescarni:
@Sceki my main takeaway from the review is that you should try to reduce the number of times you create new vectors inside the inner loops of the algo. Please try to pre-allocate vectors outside the loops and re-use them for multiple loop iterations.

There's a few additional comments and requests for small fixes as well.

include/pagmo/algorithms/gaco.hpp
darioizzo: Indeed you should only be verbose if verbosity is at least 1 ... no screen output should be produced when verbosity is zero. So leave it, but only if the verbosity is set

include/pagmo/algorithms/gaco.hpp
Sceki: @bluescarni 
I have corrected this aspect in all the abovementioned vectors, thanks. However, this `saved_value_position` is only generated once per generation when the function is called. Shall I declare it in the `main` (outside the generations loop) and pass it as an input for the function every time?

bluescarni: I think it's fine to leave it there for the time being.

include/pagmo/algorithms/gaco.hpp
Sceki: I just read why it is useful to couple `&` with `const` in these cases (here: https://stackoverflow.com/questions/44795308/why-put-const-in-c).. interesting!! thanks

darioizzo:
whats the status here? a few comments are still not replied to?

Sceki:
> whats the status here? a few comments are still not replied to?

Hello @darioizzo 

I did not reply to every comment but I did commit the changes several days ago

darioizzo:
Indeed its a bit difficult for me to check what happened, for example to the serialization remark ....and the ones above .... its always better to leave a short reply.

Anyhow, if all the comments are replied to, can you merge the current esa/master into your branch then we will merge this and start a new phase?

include/pagmo/algorithms/gaco.hpp
Sceki: @darioizzo 
Indeed there was a missing data member (`m_sol_archive`) and I added it  in the commit.

Sceki:
Ok, I just merged the current esa/master and pushed.. apparently the ci/circleci is complaining though

darioizzo:
wait, I am getting confused ... what is circle CI doing here? I thought @bluescarni still did not merge the PR on the use of circle CI #266

darioizzo:
Ok, i think its just that the circle CI is now active on all PRs and since in this one we do not have a .circleci file it complains. SHould be ok.

@bluescarni For me this is OK for a merge, I will let you the "honour" to press the button :) (circleCI will fail though and the two osx builds)

darioizzo:
@Sceki We need to discuss the way forward .... we can 
* 1) work on the multi objective extension
* 2) start working on the udp
Let me have your thoughts ... maybe via mail.

bluescarni:
@Sceki please re-merge master into the PR, so you can get the circleci support.

Sceki:
> @Sceki please re-merge master into the PR, so you can get the circleci support.

Done!

PR #253, opened 2019-03-05
Rework the rng logic in generic.hpp

bluescarni:
This PR mainly reworks the ``random_decision_vector()`` function to take in input a problem, rather than bounds, integer dimensions, etc. This simplifies the API and avoids needlessly repeating various sanity checks.

I also added a couple of getters in problem to fetch directly references to the lower/upper bounds (rather than having to create a copy of them each time ``get_bounds()`` is called) and fixed a bunch of smaller test/doc issues.


bluescarni:
@darioizzo Can you please check the modifications I did to the IHS, NSGA2 and SA algorithms? They seem simple enough and the tests pass, but would be good to get a look from you.

PR #255, opened 2019-03-09
batch_random_decision_vector()

bluescarni:
The batch-mode counterpart to ``random_decision_vector()``, for use in #226.

### DIFF
--- a/include/pagmo/utils/generic.hpp
+++ b/include/pagmo/utils/generic.hpp
@@ -37,6 +37,7 @@ see https://www.gnu.org/licenses/. */
 #include <stdexcept>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include <boost/numeric/conversion/cast.hpp>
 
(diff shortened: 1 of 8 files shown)


include/pagmo/utils/generic.hpp
```diff
@@ -310,6 +310,100 @@ inline vector_double random_decision_vector(const problem &prob, Rng &r_engine)
...
+    // LCOV_EXCL_START
+    if (n > std::numeric_limits<vector_double::size_type>::max() / nx) {
+        pagmo_throw(std::overflow_error,
+                    "Cannot generate " + std::to_string(n)
+                        + " random decision vectors in batch mode, as that would result in an overflow error");
+    }
+    // LCOV_EXCL_STOP
+
+    // Check the problem bounds, and prepare the random distributions.
+    std::vector<std::uniform_real_distribution<double>> v_rdist;
```
darioizzo: It is weird some how to be forced to create these large vectors of distribution. Would it not be somehow more efficient to use the distribution parameter setters (https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution/param, https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/param)

then you only need to construct one uniform_int and one uniform_real distribution and change the param on the fly.

bluescarni:
@darioizzo ok to merge?

darioizzo:
beautiful! Yes I merge this.

PR #265, opened 2019-04-08
Fix #256: Add basic C++ tutorial and nsga2 example.

### LINKED ISSUE #256: NSGA-II C++ example
Hi,

I'm trying to use C++ pagmo for multi-objective optimization. I can compile the single-objective example in the tutorial but I cannot compile a multi-objective one. This is what I tried:

```c++
#include <iostream>

#include <pagmo/algorithm.hpp>
#include <pagmo/algorithms/nsga2.hpp>
#include <pagmo/archipelago.hpp>
#include <pagmo/problem.hpp>
#include <pagmo/problems/dtlz.hpp>

using namespace pagmo;

int main()
{
    // 1 - Instantiate a pagmo problem constructing it from a UDP
    // (user defined problem).
    problem prob{dtlz(1, 10, 2)};

    // 2 - Instantiate a pagmo algorithm
    algorithm algo{nsga2(100)};

    // 3 - Instantiate an archipelago with 16 islands having each 20 individuals
    //archipelago archi{16, algo, prob, 20};

    // 4 - Run the evolution in parallel on the 16 separate islands 10 times.
    algo.evolve(10);

    // 5 - Wait for the evolutions to be finished
    algo.wait_check();

    // 6 - Print the fitness of the best solution in each island
    for (const auto &isl : algo) {
        for (const auto &sol : isl.get_population().get_f()) {
            std::cout << sol[0] << ' ' << sol[1] << '\n';
        }
    }
}
```

but it doesn't even compile. What am I doing wrong? It would be very helpful to have a MO example in the tutorial

Thanks,

Manuel.

MLopez-Ibanez:


doc/sphinx/docs/cpp/tutorials/cpp_tut.rst
```diff
@@ -0,0 +1,22 @@
+.. _cpp_tutorials:
+
+C++ tutorials
+================
+
+Basics
+^^^^^^
+
+After following the :ref:`install` you will be able to compile and run C++ pagmo
```
darioizzo: I think here we should use the same format of the python tutorials, where after a generic introduction on how to use the tutorials (which for the c++ case could include the compilation line for gcc), we give a list of tutorials, each on a separate page.

This would ensure that we can build and add tutorials in the future without getting a long one pager.

bluescarni:
@MLopez-Ibanez it seems like the OSX builds got stuck, I just restarted them manually.

doc/sphinx/docs/cpp/tutorials/cpp_tut.rst
MLopez-Ibanez: Sure. But right now there is only this example. Do you want to move this example to its own file? I'm not sure if I'll have time to add additional tutorials, but maybe this could be the start point for others to start writing them.

darioizzo: Indeed thats the idea .... other user can then make PRs adding tutorials easily

darioizzo:
Thanks, CI failiures are on our side, will ignore those.

PR #266, opened 2019-04-12
(WIP) CircleCI

bluescarni:
Starting to move some CI builds from travis to circleci. Advantages:

- more modern Ubuntu versions (Bionic & Cosmic), thus more stable build environment (can use directly the distro packages),
- better build performance,
- CI parallelism with Travis.

We still need Travis for the following:

- builds with the ancient GCC 4.8 (until we switch to C++>11, at least),
- OSX builds (which are not gratis on circleci).

Fixed also a couple of warnings from modern clang versions, and an ipopt quirk when using Ubuntu/Debian packages.


darioizzo:
wow!! this look super cool!!

bluescarni:
@darioizzo So it turns out the compile timeouts on travis OSX where due to the cec2013/cec2014 problems. We disabled those on OSX on certain early versions of Xcode, but the heuristic started failing after conda began to install its own compiler.

Bottom line, I have disabled unconditionally the cec2013/cec2014 problems when on OSX. We can probably re-enable them when we switch to the compiled pagmo library.

bluescarni:
@darioizzo I haven't moved over the pygmo builds to circleci yet, but perhaps it's better to merge this PR immediately (after CI passes) so that we can unblock the build failures on the other PRs.

PR #268, opened 2019-04-18
add gwo for cpp

michiboo:
Hi! So I am making my first PR here! 
I have coded a implementation for grey wolf optimizer mainly based on the python implementation from https://github.com/7ossam81/EvoloPy/blob/master/GWO.py. 

I would like a code review before I proceed any further, to making sure I am in the right direction!
one question that I have is should I return the population or only the alpha position in the evolve method?

Thank you and sorry for any mistakes in advanced!!


include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+            }
+
+        // Main gwo iterations
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) 
+        {
+             // Start of the loop through each agent
+            for (decltype(NP) i = 0u; i < NP; ++i){
+                  // clip position value that goes beyond search space          
+              for(decltype(dim) k = 0; k < dim; ++k)
+                boost::algorithm::clamp(agents_position[i][k], lb, ub);
```
darioizzo: We use these utils for out of bounds corrections since its a recurring algoritmic feature: https://github.com/esa/pagmo2/blob/master/include/pagmo/utils/generic.hpp#L513

If you want to use boost::clamp (which is a good idea) you can call it in ```force_bounds_stick```.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+
+
+                auto newfitness = prob.fitness(agents_position[i]);
+
+                // Update alpha, beta and delta
+                if (newfitness < alpha_score)
```
darioizzo: newfitness will be a ```std::vector``` as in pagmo it contains the objfuns and the constraints, so you need to access its [0] element that in single objective setups corresponds to the objective itself.

darioizzo: Clearly update also below

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+
+                if(newfitness > alpha_score && newfitness > beta_score && newfitness <  delta_score){
+                    delta_score = newfitness;
+                    delta_pos = agents_position[i];
+
+                }
+               
+        
+            }
+             a= 2-gen*((2)/m_gen);
```
darioizzo: careful here ... implicit conversions are everywhere .... 2 is int gen is unsigned m_ge too and a double .....

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+        vector_double d_vector(3); // position vector for alpha, beta and delta 
+        vector_double x_vector(3); // vector used to calculate position of an agent in a single dimension
+        double a; //coefficient which decrease linearly from 2 to 0 over generations
+        double r1, r2; // random coefficient between 0 and 1;
+        double alpha_score = std::numeric_limits<double>::infinity();
+        double beta_score = std::numeric_limits<double>::infinity();
+        double delta_score = std::numeric_limits<double>::infinity();
+        
+
+        // assume pop.get_x() return vector<vector_double>
```
darioizzo: correct

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+        auto  agents_position = pop.get_x();
+        
+
+
+        vector_double tmp(dim);                              // hold temporary value
+        std::uniform_real_distribution<double> drng(0., 1.); // to generate a number in [0, 1)
+
+        // initialize position of search agents
```
darioizzo: Here agents_position contains already the initial chromosomes, no need to initialize them inside the evolve .... also becaus by pushing back you are augmenting the population size.

include/pagmo/algorithms/gwo.hpp
darioizzo: Also I am not sure I would clip here at the begginning of the loop as we are guaranteed that the agents are valid upon entrance to the loop, why not clipping at the end?

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) 
+        {
+             // Start of the loop through each agent
+            for (decltype(NP) i = 0u; i < NP; ++i){
+                  // clip position value that goes beyond search space          
+              for(decltype(dim) k = 0; k < dim; ++k)
+                boost::algorithm::clamp(agents_position[i][k], lb, ub);
+
+
+                auto newfitness = prob.fitness(agents_position[i]);
```
darioizzo: in the first loop the ```fitness``` this is known as pop contains already the fitness values

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+    typedef std::tuple<unsigned int, double, double, double> log_line_type;
+    /// The log
+    typedef std::vector<log_line_type> log_type;
+
+    /// Constructor
+    /**
+     * Constructs a Grey Wolf Optimizer
+     *
+     * @param gen number of generations.
+     * @param agent number of search agents(wolf)
```
darioizzo: is m_agent including the alpha beta gamma? In other words pop.size() is m_agent+3 ?

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+                 agents_position.push_back(tmp);
+            }
+
+        // Main gwo iterations
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) 
+        {
+             // Start of the loop through each agent
+            for (decltype(NP) i = 0u; i < NP; ++i){
+                  // clip position value that goes beyond search space          
+              for(decltype(dim) k = 0; k < dim; ++k)
```
darioizzo: indentation is strange here 2 spaces .... did you run clang format?

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+                
+                
+                    a_vector[2] =2*a*r1-a; 
+                    c_vector[2]=2*r2;
+
+                    d_vector[2]=abs(c_vector[2]*delta_pos[k]-agents_position[i][k]); // Equation (3.5)-part 3
+                    x_vector[2]= delta_pos[k]-a_vector[2] * d_vector[2]; // Equation (3.6)-part 3          
+                
+                agents_position[i][k] = (x_vector[0]+x_vector[1]+x_vector[2])/3;  // Equation (3.7)
+
```
darioizzo: here is where I would compute the fitness (as the positions have changed) and thus update alpha beta gamma and call the ```pop.set_xf``` method that would also update the pop without causing further fitness evaluations.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+                    // Logs
+                    m_log.emplace_back(gen, alpha_score, beta_score, delta_score);
+                }
+            }
+
+        }// end main gmo iterations
+         
+        
+
+        return pop;
```
darioizzo: indeed ... but in your code you have not update it :)

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+    /**
+     * @return the seed controlling the algorithm stochastic behaviour
+     */
+    unsigned int get_seed() const
+    {
+        return m_seed;
+    }
+    /// Sets the algorithm verbosity
+    /**
+     * Sets the verbosity level of the screen output and of the
```
darioizzo: docs to be updated

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+    /**
+     * Constructs a Grey Wolf Optimizer
+     *
+     * @param gen number of generations.
+     * @param agent number of search agents(wolf)
+     * @param seed seed used by the internal random number generator (default is random)
+     *
+     */
+   
+        gwo(unsigned int gen = 1u, unsigned int seed = pagmo::random_device::next())
```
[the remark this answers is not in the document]
michiboo: I was thinking that pop.size is same as agent_size. Also it is better to follow naming from other pagmo2 algorithms than using unique name such as agent?

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,351 @@
...
+        }
+        if (prob.is_stochastic()) {
+            pagmo_throw(std::invalid_argument,
+                        "The problem appears to be stochastic " + get_name() + " cannot deal with it");
+        }
+        // Get out if there is nothing to do.
+        if (m_gen == 0u) {
+            return pop;
+        }
+        if (pop.size() < 5u) {
```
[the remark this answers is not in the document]
michiboo: I have changed to 3 only as it require alpha, beta, delta at minimum

include/pagmo/algorithms/gwo.hpp
michiboo: Agree

include/pagmo/algorithms/gwo.hpp
michiboo: removed m_agent as it is the same as pop.size()

include/pagmo/algorithms/gwo.hpp
michiboo: I try to compile the quick example with gwo and found it come up with type error that it require return type population, is it a convention of pagmo2?

michiboo:
Hi, so it can finally compile successfully! I guess next step will be testing and python binding? Could you provide some direction on these two task?

Thanks!

include/pagmo/algorithms/gwo.hpp
darioizzo: Yes the fitness method must return the evolved pop ... that's why I mentioned you must call the pop.set_xf method to insert an evolved individual into the pop

darioizzo:
For testing have a look at some other test .... like the ones for psp or de. In general we test so that each line of code is called and all errors are triggered. (Unit test) since it's tricky to test the correctness of the evolve method .... there we only test for errors and feature like determinism of seed control and log creation. The best to do is follow another test. 

As for the correctness of the evolve method, just try to check if on some problems the performances reported in literature are met.

michiboo:
Hi I have completed the test!

darioizzo:
Can you check and fix the ci failiures?

michiboo:
Hi it is fixed! I want to ask what's the next step for gwo? or I can add another optimization algorithm such as the whale optimization algorithm.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,364 @@
...
+     * This method will save/load \p this into the archive \p ar.
+     *
+     * @param ar target archive.
+     *
+     * @throws unspecified any exception thrown by the serialization of the UDP and of primitive types.
+     */
+    template <typename Archive>
+    void serialize(Archive &ar)
+    {
+        ar(m_gen, m_seed, m_verbosity, m_log);
```
bluescarni: The ``m_e`` member should be serialised as well.

bluescarni:
@michiboo you also need to add the test to the ``CMakeLists.txt`` file in the ``test`` subdirectory, and add the ``gwo.hpp`` header to the list of headers in ``pagmo.hpp``.

bluescarni:
@michiboo also, you should add the new algorithms to the documentation files for C++:

https://github.com/esa/pagmo2/tree/master/doc/sphinx/docs/cpp/algorithms

Similar things for the python side, when you have exposed it:

https://github.com/esa/pagmo2/blob/master/doc/sphinx/docs/python/algorithms/py_algorithms.rst

You should also add the new algorithm to the global list of algorithms here:

https://github.com/esa/pagmo2/blob/master/doc/sphinx/docs/algorithm_list.rst

bluescarni:
@michiboo new algorithms are always welcome! But they should be done in separate PRs.

bluescarni:
@michiboo  oh and if you feel like it, you could write down the steps needed to add a new pagmo algortihm in a wiki page :)

michiboo:
@bluescarni thanks for the info! adding algorithm wiki will be the next thing to do after this PR is finish! 
Do you have any example file where you bind the algorithm to python?

bluescarni:
> @bluescarni thanks for the info! adding algorithm wiki will be the next thing to do after this PR is finish!
> Do you have any example file where you bind the algorithm to python?

Great!

For the Python exposition, you can take a look at this file:

https://github.com/esa/pagmo2/blob/master/pygmo/expose_algorithms_1.cpp

You can see how it is done for, e.g., PSO, and add the exposition code for GWO at the bottom of that file. You will also need to provide appropriate Python documentation strings (docstrings) for the GWO class and its methods. The docstrings are to be added to the ``docstrings.hpp`` and ``docstrings.cpp`` files.

bluescarni:
@michiboo good job! It's looking good to me, perhaps @darioizzo wants to comment as well?

The travis docs build is failing because of some issue in the format of the documentation:

https://travis-ci.org/esa/pagmo2/jobs/523113960#L1983

Should be fairly easy to fix, but feel free to ask if you need help.

bluescarni:
@michiboo the travis failure message is a bit garbled and repeated a few times, but it essentially boils down to this:

```
/home/travis/build/esa/pagmo2/doc/sphinx/docs/cpp/algorithms/gwo.rst:2: WARNING: Content block expected for the "seealso" directive; none found.
/home/travis/build/esa/pagmo2/doc/sphinx/docs/cpp/algorithms/gwo.rst:3: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/travis/build/esa/pagmo2/doc/sphinx/docs/cpp/algorithms/gwo.rst: WARNING: document isn'\''t included in any toctree
```

The first 2 warnings are referring to the C++ documentation (the one immediately preceding the gwo algorithm's source code in ``gwo.hpp``, which is then parsed by doxygen and included in the sphinx docs via the breathe extension). It seems like there might be some indentation/formatting errors, in relation to the ``seealso`` directive as well.

The third warning is sphinx complaining that the ``gwo.rst`` document has not been included in any index/table of contents. When a new algorithm's docs are added, the file should be added to the algorithms TOC here:

https://github.com/esa/pagmo2/blob/master/doc/sphinx/docs/cpp/cpp_docs.rst

bluescarni:
@michiboo I also spoke briefly to @darioizzo earlier, who is traveling at the moment. He mentioned he would like to see a preliminary assessment of the algorithm's performance before proceeding with the merge of this PR.

Perhaps you could look into doing a tutorial similar to this one for ``gwo``?

https://esa.github.io/pagmo2/docs/python/tutorials/cec2006_gaco_benchmark.html

michiboo:
Hi @bluescarni  I have create a example of using gwo, the solution seem okay for me. It would be great if you can run it and test if it is really OK! 
you can find the file here https://gist.github.com/michiboo/2df766cd7f59c0eb579290c8cbf2536e

bluescarni:
The issue with the doc formatting is still there. The problem is that the contents of the ``seealso`` directive are not indented as they should:

https://github.com/esa/pagmo2/pull/268/files#diff-1ca978818fb80eacc38af5af3a463b0bR54

I.e., the contents of the ``seealso`` section need to be indented by three spaces (see how it is done elsewhere in pagmo).

bluescarni:
@michiboo can you also post the graphs that are outputted by the benchmark you posted above?

michiboo:
@bluescarni I have added the graph to the link!

bluescarni:
@michiboo Thank you!

michiboo:
Hi I think all build are successful but there is one that has "HTTP 000 CONNECTION FAILED", it should be pass when retest it again. Is it good idea to squash all commit into one?

bluescarni:
> Hi I think all build are successful but there is one that has "HTTP 000 CONNECTION FAILED", it should be pass when retest it again. Is it good idea to squash all commit into one?

Yes the HTTP error is transient and can be safely ignored.

I am neutral about the commit squashing, but before merging the PR we should wait until @darioizzo comes back from travelling so he can have a final look.

If you want to start working on something else (I think you mentioned another algorithm you would like to implement?), please feel free to open a new PR so that the work can proceed in parallel while we wait.

michiboo:
I am working on a step-by-step guide for adding new algorithm!

bluescarni:
> I am working on a step-by-step guide for adding new algorithm!

Great, thanks!

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,362 @@
...
+        double delta_score = std::numeric_limits<double>::infinity();
+
+        // assume pop.get_x() return vector<vector_double>
+        auto agents_position = pop.get_x();
+        std::uniform_real_distribution<double> drng(0., 1.); // to generate a number in [0, 1)
+
+        // Main gwo iterations
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) {
+
+            auto fit = pop.get_f();
```
darioizzo: probably better to use ```pop.get_f()[i]``` here?

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,362 @@
...
+    /**
+     * Sets the verbosity level of the screen output and of the
+     * log returned by get_log(). \p level can be:
+     * - 0: no verbosity
+     * - >0: will print and log one line each \p level generations.
+     *
+     * Example (verbosity 1):
+     * @code{.unparsed}
+     *  Gen:         Alpha:          Beta:         Delta:
+     * 1         5.7861        12.7206        19.6594
```
darioizzo: fix indentation

include/pagmo/algorithms/gwo.hpp
michiboo: To my understanding, pop.get_f() return const std::vector<vector_double>, which contain a vector_double for each individual in population, in line 165 I index it  by using i and get the 0 index of this vector which allow me to get the fitness of individual i. So If I changed it to pop.get_f()[i] in 159, it should not work because I will then only index into the first individual in the population? please correct me if I'm wrong!

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,362 @@
...
+        double alpha_score = std::numeric_limits<double>::infinity();
+        double beta_score = std::numeric_limits<double>::infinity();
+        double delta_score = std::numeric_limits<double>::infinity();
+
+        // assume pop.get_x() return vector<vector_double>
+        auto agents_position = pop.get_x();
+        std::uniform_real_distribution<double> drng(0., 1.); // to generate a number in [0, 1)
+
+        // Main gwo iterations
+        for (decltype(m_gen) gen = 1u; gen <= m_gen; ++gen) {
```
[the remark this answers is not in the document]
michiboo: Hi.. getting the best three individuals at the initialization, in the first part main gwo iterations from line 164 to line 178 it is already done here...
if we initialize it with best three individuals before gwo iteration, these code will run for nothing or we add a condition that it shouldn't run in the first iteration which is unnecessary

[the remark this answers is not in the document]
darioizzo: The way its currently written is incorrect. Upon entering the main loop alpha beta and gamma need to be defined and not contain zeros and infinities. Otherwise on line 198, for example, and during the first iteration (i=0), beta will contain garbage.

I agree that the update of alpha beta and gamma can be done later in the loop, after line 214.

include/pagmo/algorithms/gwo.hpp
darioizzo: true sorry did not see that this was outside the loop.

include/pagmo/algorithms/gwo.hpp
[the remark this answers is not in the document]
michiboo: You are right, the performance of the algorithm seem to be increase after this changes!

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,365 @@
...
+        vector_double a_vector(3);          // vector coefficient for encircling prey
+        vector_double c_vector(3);          // vector coefficient for encircling prey
+        vector_double d_vector(3);          // position vector for alpha, beta and delta
+        vector_double x_vector(3);          // vector used to calculate position of an agent in a single dimension
+        double a;                           // coefficient which decrease linearly from 2 to 0 over generations
+        double r1, r2;                      // random coefficient between 0 and 1;
+        std::vector<std::size_t> index_vec; // used to stored sorted index
+        auto agents_position = pop.get_x();
+        auto init_fit = pop.get_f();
+        for (unsigned int i = 0; i != init_fit.size(); ++i)
```
darioizzo: use ```std::iota```.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -0,0 +1,365 @@
...
+            index_vec.push_back(i);
+        std::sort(index_vec.begin(), index_vec.end(),
```
darioizzo: This is UB (Undefined Behaviour) whenever the fitness returned by the UDP is NaN or Inf (which is not preventable). In this case the `<` operator is not meeting the compare requirments (https://en.cppreference.com/w/cpp/named_req/Compare)

In Pagmo we thus use https://github.com/esa/pagmo2/blob/master/include/pagmo/detail/custom_comparisons.hpp#L48

darioizzo:
Nice work!! After you fix these last small things can you reproduce the plots in the gist above and also add some of the problems in the original paper, taking care to check that the performances are in line with what declared there?

Then we are good to go.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -136,19 +138,19 @@ class gwo
...
+        vector_double x_vector(3); // vector used to calculate position of an agent in a single dimension
+        double a;                  // coefficient which decrease linearly from 2 to 0 over generations
+        double r1, r2;             // random coefficient between 0 and 1;
         auto agents_position = pop.get_x();
         auto init_fit = pop.get_f();
-        for (unsigned int i = 0; i != init_fit.size(); ++i)
-            index_vec.push_back(i);
-        std::sort(index_vec.begin(), index_vec.end(),
-                  [&](unsigned int a, unsigned int b) { return init_fit[a][0] > init_fit[b][0]; });
+        std::vector<unsigned int> index_vec(init_fit.size()); // used to stored sorted index
```
bluescarni: @michiboo this is a vector of indices into the ``init_fit`` vector, thus it should be a ``std::vector<decltype(init_fit.size())>`` rather than a ``std::vector<unsigned>``.

include/pagmo/algorithms/gwo.hpp
```diff
@@ -136,19 +138,19 @@ class gwo
...
+        std::iota(index_vec.begin(), index_vec.end(), 0);
```
bluescarni: @michiboo when you use ``iota``, you must make sure that the initial value, 0 in this case, is of the correct type. The way you are using it here, the result of ``iota`` will be a signed ``int``, which is not what you want. You should write ``decltype(init_fit.size())(0)`` instead of just ``0``.

include/pagmo/algorithms/gwo.hpp
michiboo: Thanks for reviewing!

michiboo:
I have question about adding problems from paper, I notice some of the problem are only implement in problems suite. Does that mean I cannot customize it? For example Sphere Function is in only CEC2013 but not in Scalable problems. should I add sphere function to scalable problem? or I should create additional problems suite for gwo.

bluescarni:
I am not familiar with the GWO paper, but it seems to me it is referring both to the CEC2005 problems and classical test problems. I don't think it is necessary to reproduce all the results from the paper, but it would certainly be handy to have the cec2005 problems in pagmo.

Maybe what you could do for this PR is to reproduce a few  results from the paper without adding the optimisation problems to pagmo, and leave that for a separate PR if/when you or someone else is willing to implement it. But let's hear from @darioizzo first.

darioizzo:
Yes, indeed we only need some indication we have correctly interpreted the paper original algo we refer to .... so a few quick checks would be enough ....

michiboo:
Hi I have updated the gist link  https://gist.github.com/michiboo/2df766cd7f59c0eb579290c8cbf2536e,
I used some of the classical test function from pagmo2 and from CEC2013 and compare it with DE and PSO. The result look acceptable to me.

darioizzo:
Good to go if CI is green.

darioizzo:
@michiboo 
Thank you very much for the PR, very nice work and quickly done....

PR #279, opened 2019-05-06
Added installation instructions for ArchLinux and FreeBSD, organized installation page slightly

acxz:


bluescarni:
Thanks for the PR! I have restarted some travis builds which failed due to transient network errors.

doc/sphinx/install.rst
```diff
@@ -55,6 +61,55 @@ and then we can immediately install pagmo:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Pagmo is also available on the `Arch User Repository
+<https://aur.archlinux.org>`__ (AUR) in Arch Linux. It is
+recommended to use an AUR helper like
+`yay <https://aur.archlinux.org/packages/yay/`__ or
+`pikaur <https://aur.archlinux.org/packages/pikaur/`__ for ease of installation.
+See the `AUR helpers <https://wiki.archlinux.org/index.php/AUR_helpers` page on
+the Arch Linux wiki for more info.
+
+Note: To install pagmo with optional dependency support like nlopt or ipopt,
```
bluescarni: I have a question about this. Wouldn't it be better to have pagmo depend on all optional packages instead and enable them unconditionally in the build script for Arch linux?

acxz: I see the reasoning behind that, but I don't want users to install packages that they have no need for. Let's say I have a package which has 100 optional dependencies, but I only need to use 1 of them. I don't want the hassle of the rest of the 99 packages. Rather I should build only based on what the current system has. They are optional for a reason and I would like to keep it that way, unless if the are required by the CMakelists.txt from the source.

bluescarni: Ok, fair enough!

bluescarni:
@acxz looks good to me. Can I merge or would you like to add something else?

acxz:
I think it's good to go.

PR #281, opened 2019-05-07
Fix building without Eigen support

jslee02:
`xnes.hpp` should be included only when Eigen support is enabled. Otherwise, the compiler complains by [this line](https://github.com/esa/pagmo2/blob/master/include/pagmo/algorithms/xnes.hpp#L588).

Related issue: https://github.com/dartsim/dart/issues/1310


bluescarni:
Absolutely correct, thanks for catching that! Gonna go ahead and merge, the build failure on appveyor is a temporary network error.

PR #289, opened 2019-05-12
gaco with bfe

Sceki:
addition of recently implemented batch function evaluation scheme (`bfe`) in `gaco` algorithm

### DIFF
--- a/tests/gaco.cpp
+++ b/tests/gaco.cpp
@@ -1,30 +1,30 @@
 /* Copyright 2017-2018 PaGMO development team
-
-This file is part of the PaGMO library.
-
-The PaGMO library is free software; you can redistribute it and/or modify
-it under the terms of either:
-
-  * the GNU Lesser General Public License as published by the Free
-    Software Foundation; either version 3 of the License, or (at your
-    option) any later version.
-
-or
-
-  * the GNU General Public License as published by the Free Software
-    Foundation; either version 3 of the License, or (at your option) any
-    later version.
-
-or both in parallel, as here.
-
-The PaGMO library is distributed in the hope that it will be useful, but
-WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
-or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
-for more details.
-
-You should have received copies of the GNU General Public License and the
-GNU Lesser General Public License along with the PaGMO library.  If not,
-see https://www.gnu.org/licenses/. */
+ 
+ This file is part of the PaGMO library.
+ 
+ The PaGMO library is free software; you can redistribute it and/or modify
+ it under the terms of either:
+ 
+ * the GNU Lesser General Public License as published by the Free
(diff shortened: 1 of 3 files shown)


tests/gaco.cpp
```diff
@@ -343,3 +345,46 @@ BOOST_AUTO_TEST_CASE(integer_test_2)
...
+    auto start_time_2 = std::chrono::high_resolution_clock::now();
+    population pop_2{lennard_jones{40u}, 200u, 23u};
+    gaco uda_2{10u, 10u, 1.0, 25.0, 0.01, 5u, 7u, 1000u, 1000u, 0.0, 10u, 0.9, false, 23u};
+    uda_2.set_seed(23u);
+    pop_2 = uda_2.evolve(pop_2);
+    auto end_time_2 = std::chrono::high_resolution_clock::now();
+    std::cout << "gaco on Lennard Jones without bfe duration:" << std::endl;
+    std::cout << std::chrono::duration_cast<std::chrono::seconds>(end_time_2 - start_time_2).count() << ":";
+    std::cout << std::chrono::duration_cast<std::chrono::microseconds>(end_time_2 - start_time_2).count();
+}
```
Sceki: @bluescarni, @darioizzo 
lacking other examples on how to implement the test with `bfe`, I thought about adding a test in which the time with and w/o `bfe` is printed to screen in case a large population size is chosen, to actually show that the two times differ..  I could have done a check for the time (showing that with `bfe` it was indeed smaller), but then I do not think that this would be verified in all the machines..

I am not sure this is what you actually want .. I can change it as you wish or remove it in case it is not appropriate.

Sceki: PS. I just saw the first CI failure due to the lack of the 'b' when documenting `set_bfe`.. I will correct it in the next commit

darioizzo:
Thanks for the PR at a glance its going in the direction we discussed .... Unfortunatley its impossible to review this PR as it appears you have some weird setting in the editor, which then commits changes to spaces and other formatting things. As a consequence it seems you have  changed every line of the code (see the diff).

Can you figure this out and resubmit a PR that is clean?

In case you want to know a good editor for developing, visual studio code is what I have been using lately ....

bluescarni:
@Sceki we tend to use visual studio code, but it's not mandatory to do so. The important bit is that all of pagmo's code must be formatted with ``clang-format`` following the style file provided in the source of pagmo's source tree (and called ``.clang-format`` with a leading dot). Many editors support ``clang-format`` either natively or via an extension/plugin.

tests/gaco.cpp
bluescarni: @Sceki unit tests are supposed to test the correctness of the code, so an appropriate test would be to check whether two evolutions with identical initial conditions (including random seeds), but one with ``bfe`` and the other without,  produce identical results.

Sceki:
Ok, I completely forgot to clang-format the `.cpp` files, my bad.. just created the other PR, gonna close this!

PR #293, opened 2019-05-14
Manylinux attempt

darioizzo:
Here we use the dockerhub image to build pagmo wheels. Main reasons are:

* speed
* ipopt is now included


bluescarni:
:+1: :fireworks:

PR #294, opened 2019-05-15
C++ improvements

bluescarni:
This PR adds a couple of capabilities on the C++ side:

- generic assignment of a UDP/UDA/... to a ``problem``/``algorithm``/..., so that now it is possible to write:

```c++
problem p;
p = rosenbrock{};
```

- a new function ``is_valid()`` for user-defined classes, with which it is possible to query if the object was moved from:

```c++
problem p;
assert(p.is_valid());
problem p2 = std::move(p);
assert(!p.is_valid());
```

We already stated in the documentation that moved-from objects could only be destroyed and assigned to (revived). Now it is also possible to call ``is_valid()`` to query the current state of the object.

None of these features are meaningful in Python, so pygmo was not touched by these additions.

I also moved ``null_problem`` and ``null_algorithm`` to their own headers (they previously resided in ``problem.hpp`` and ``algorithm.hpp``), and made some other smaller internal changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,7 @@ if(PAGMO_BUILD_PAGMO)
         "${CMAKE_CURRENT_SOURCE_DIR}/src/rng.cpp"
         "${CMAKE_CURRENT_SOURCE_DIR}/src/threading.cpp"
         # UDP.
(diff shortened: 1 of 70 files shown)


darioizzo:
Looks good to me... happy about the null prob algo move !!

PR #302, opened 2019-06-09
bug fix for gaco

Sceki:
In the algo, in an if cycle, it was only checked one element of a vector, instead of the whole vector. 
This is a bug, since there are cases in which two vectors have the same initial element (in this case penalty) but they are different, and the algo was not able to recognize that. 
This was not easy to spot at first since, in general, each ant has a different penalty (and thus the first element of the vector is indeed different): but this is not always true, and there might be cases in which the penalty is the same for two different ants.


bluescarni:
@Sceki thanks for the PR! I'll fix the CI issue in a separate PR.

PR #303, opened 2019-06-09
Various small fixes

bluescarni:


bluescarni:
@Sceki I had to touch some code in wfg and gaco in order to quench some compiler warnings. Would you please take a look in order to make sure I did not accidentally wreck anything?

In wfg, it was mostly a matter of casting some integral quantities to double. In gaco, it seems like there are some inconsistencies in the use of integral types (such as, using an ``unsigned`` for counting fevals and then adding an ``unsigned long`` to it, and another similar occurrence). I just used static casts for now, but it would be good to understand if we can uniform the integral types used instead of suppressing the warnings with casts.

src/problems/wfg.cpp
```diff
@@ -133,7 +131,7 @@ std::pair<vector_double, vector_double> wfg::get_bounds() const
 {
     vector_double upper_bounds(m_dim_dvs);
     for (decltype(m_dim_dvs) i = 0u; i < m_dim_dvs; ++i) {
-        upper_bounds[i] = 2.0 * (i + 1);
+        upper_bounds[i] = 2.0 * (static_cast<double>(i) + 1);
```
Sceki: @bluescarni how would you handle such a case (in which the `i` is used for both accessing the vector and also in assigning its element), if not with a cast?

Sceki:
Hi, thanks for the corrections, and sorry for the burden, I should have paid more attention. 
I will correct all of those in a separate PR when this will be merged if you agree. 
My only doubt is what to do in the case that I mentioned in the above comment (which happens a lot in `wfg`).
Also, how come I could not see these errors when compiling in debug mode?

src/problems/wfg.cpp
bluescarni: @Sceki sorry I was not clear. The cast to double here is ok, as you want to convert that ``i`` to a floating-point anyway as it makes sense for the computation. In this case, I believe GCC is being a bit too strict, as it is complaining that the implicit cast to double might not preserve the exact value of the original variable (you are converting a 64bit integer to double precision, so if the number has a magnitude larger than 2**53 the conversion will not be exact - that's why GCC wants the static cast, to make extra sure that you are ok with the potential loss of precision). 

So keep the cast here, even if it looks ugly.

bluescarni:
> Also, how come I could not see these errors when compiling in debug mode?

These result only in warnings and only on GCC (not on clang/MSVC). When compiling with GCC, we don't treat warnings as errors, we do that only with Clang builds  as GCC's warnings tend to be a little less useful than Clang warnings currently (it often has false positives, etc.).

bluescarni:
@Sceki I just realised there's a bunch of additional warnings in wfg:

https://ci.appveyor.com/project/ci4esa/pagmo2/builds/25153063/job/a1v560etgjfntoke#L254

These show up only on Windows because on that platform ``unsigned long`` and ``unsigned long long`` have different ranges - 32bit vs 64bit - whereas on Linux both ``unsigned long `` and ``unsigned long long`` are 64bit. If you could take care of these, that would be great (ignore the other warning about the ``pure`` function attribute).

Sceki:
Ok, as soon as this will be merged I will do a separate PR then!

PR #304, opened 2019-06-09
Following from PR #303

Sceki:
To make the types uniform in `gaco` algorithm and `wfg` problem, as pointed out in #303


bluescarni:
@Sceki looks good to me, thanks for taking care of this!

PR #307, opened 2019-06-11
(WIP) #302 Correction 

Sceki:
I am not convinced that #302 fixed the bug, I am going to keep this open until I am sure. I will rework and clean the `update_sol_archive` function so that everything is more efficient and possibly more recognizably bug free


Sceki:
@darioizzo , @bluescarni 
For me (if the tests pass), this is ready. I have re-worked a bit the logic behind it, and I made sure that the bug is fixed also in this case. I have manually checked the `sol_archive` update mechanism, and it seems to work fine.
Please, let me know if you see things to correct or errors or inefficiencies.

PS: I had to use a `push_back` because I could not foresee the size of a vector (there was also before).

darioizzo:
Had a quick look and test, seems ok to me ...

PR #314, opened 2019-06-26
NSPSO

Sceki:
This PR aims to port Non-dominated Sorting Particle Swarm Optimizer (NSPSO) `uda` from old pagmo.


tests/nspso.cpp
```diff
@@ -30,6 +30,19 @@ see https://www.gnu.org/licenses/. */
...
+#include <limits>
+#include <numeric>
+#include <random>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
```
Sceki: I forgot to erase these includes.. in the next commit I will do it

src/algorithms/nspso.cpp
```diff
@@ -141,12 +141,10 @@ population nspso::evolve(population pop) const
         maxv[i] = vwidth;
     }
     // Initialize the particle velocities if necessary
-    if ((m_velocity.size() != swarm_size)) { // if with memory, add: || (!m_memory)
-        m_velocity = std::vector<vector_double>(swarm_size, dummy_vel);
-        for (decltype(swarm_size) i = 0u; i < swarm_size; ++i) {
-            for (decltype(n_x) j = 0u; j < n_x; ++j) {
-                m_velocity[i][j] = uniform_real_from_range(minv[j], maxv[j], m_e);
-            }
+    m_velocity = std::vector<vector_double>(swarm_size, dummy_vel);
```
Sceki: Just for clarity.. this (I believe) was causing the issue.. the line: `(m_velocity.size() != swarm_size` should always be false in my tests, but for some test cases it resulted to be true, thus causing the `m_velocity` not to be re-initialized for some cases, and thus it resulted to have the wrong size.. tomorrow I will investigate better why this happened (if I confirm that this is indeed the issue)

bluescarni: Is this coming from the original nspso implementation?

Sceki: As far as I understood, in the previous pagmo the velocity was generated random in the first individual (similar to the decision vector) but in this pagmo is not possible anymore, so I have decided to do the same as it is done in pso.hpp to initialize the velocity

Sceki: In any case, I will have to go through this part again anyway when I will introduce memory

src/algorithms/nspso.cpp
```diff
@@ -0,0 +1,526 @@
...
+            }
+
+            best_non_dom_indices = std::vector<vector_double::size_type>(
+                best_non_dom_indices.begin(),
+                best_non_dom_indices.begin()
+                    + static_cast<vector_double::difference_type>(i)); // keep just the non dominated
+        }
+
+        // Decrease w from m_max_w to m_min_w troughout the run
+        double w = m_max_w - (m_max_w - m_min_w) / m_gen * gen;
```
Sceki: @darioizzo @bluescarni 
From my understanding, due to this line, I cannot implement `memory` in this algorithm.. the problem is that when memory is inserted, there is no way for me to know apriori the total number of generations (`m_gen`), although I could count the `gen`.

darioizzo: @sceki Indeed. These are shitty heuristic that I am surprised people still use. Defining a schedule for w is ok, but having it depend on m_gen/gen is horrible. Is this in the original paper too?

Sceki: @darioizzo  I actually do not think so. I believe that the authors have decided this arbitrarily. In one of the three papers that they mention as reference ( https://titan.csit.rmit.edu.au/~e46507/publications/nspso-gecco03.pdf ), it is said that `w` is gradually decreased from 1.0 to 0.4 .. not sure how they achieved this decrease (which probably also makes the results difficult to reproduce)

darioizzo: Would it then make sense to leave it fixed? The user could then implement a schedule on ```w``` manually. something like:
```
pop = algo.evolve(pop)
algo.extract(nspso).set_w(2u)
```

Sceki: Sure. This seems more meaningful.

This could actually also be done in the normal pso, in which, right now, the weight is only passed as input ( and it is thus fixed )

darioizzo: If w is fixed do you still need memory in this algorithm? Are there data members that change during the evolve?

Sceki: Sorry but right now I’m not responsive.. I’ll be operative again and take care of this from the 16th!

darioizzo: ok .....

Sceki: > If w is fixed do you still need memory in this algorithm? Are there data members that change during the evolve?

If I understood well the question yes. There are other variables that would change when memory is implemented, and thus require a workaround if memory wants to be introduced (should not be difficult though). Shall I make `w` fixed and chosen by the user and also modify the algorithm in such a way that memory is introduced? 
Or shall I just do the first and leave the algo w/o memory?

darioizzo: Make w fixed. If the algorithm has then still some state, introduce the memory kwarg as ausual.

Final objective is that (in pseudocode, api is not respected)
```pop = my_algo(gen=100).evolve(pop)```
and
```
for i in range(100)
    pop = my_algo(gen=1, memory=True).evolve(pop)
```

darioizzo:
Do you have indication of the algorithm performance on test sets?

Sceki:
Not yet, I first wanted to discuss with you how to produce them.. Shall I use `wfg`, `zdt` or `dtlz`? Shall I only compare the final front w.r.t. `moead` and `nsga2`?

darioizzo:
Why choosing, we can easily do them all ..... just make some notebooks that you can share with us!

So, I will be leaving today for some weeks. I talked to @CoolRunning and he will be following you and basically making sure that the algorithms maco and nspso are of value as in they solve poblems and are not outperformed consistently. You can contact him on gitter or here ... in thr PR.

CoolRunning:
Hi! Let me know if you have questions about the benchmarking! I am curious about the results...

bluescarni:
@Sceki the appveyor failure seems like a transient network error. I restarted the build.

darioizzo:
@Sceki As discussed, this is ready for merge. Are you happy with it? Or it makes sense to keep it opened for a few more weeks?

Sceki:
@darioizzo  Unless you want to adjust/modify something, for me this is ready.. I left the WIP because I forgot to remove it:)

PR #319, opened 2019-07-05
Consistently disable tracking for all s11n implementations.

bluescarni:


bluescarni:
The Boost serialization library is capable of tracking by memory address the object being serialized. This can be useful because it prevents writing the same object multiple times to the same archive. It can however cause subtle issues when, for instance, (de)serializing an object via a copy or when pointers are involved.

In pagmo we already disable tracking automatically for all user-defined object (UDP, UDA, etc.). This PR disables tracking also for all other classes for which pagmo provides a serialization implementation (i.e., the type-erased containers, random number generators, Eigen matrices, etc.).

Issue #318, opened 2019-07-04
MOEAD gets stuck for popsizes = 2


Sceki:
Hello,

I seem to have encountered a bug in `moead`. The problem seems to appear when only 2 individuals are chosen as `popsize`. The following minimal example reproduces it:
```
BOOST_AUTO_TEST_CASE(minimal_example)
{
    problem prob{zdt{1u, 30u}};
    population pop1{prob, 2u, 23u};

    moead user_algo1{100u, "grid", "tchebycheff", 1u, 1., 0.5, 20., 0.9, 2u, true, 23u};
    user_algo1.set_verbosity(1u);
    pop1 = user_algo1.evolve(pop1);
}
```

bluescarni:
@Sceki any idea on where it gets stuck? I took a quick glance but I didn't see anything obvious popping up related to the number of individuals ``NP``.

Sceki:
Well, my guess is that is inside the `while` cycle in the `select_private` method.. since the algo seems to do an infinite loop.. but I am not sure.. I will do a check now

Sceki:
Indeed I confirm.. that loop is executed infinitely, thus causing the algorithm to get stuck..

bluescarni:
So is the issue the fact that it's trying to extract 2 parents from a population of 2? Would bumping up the minimum population size for MOEAD from 2 to 3 solve the issue then?

Sceki:
I am not sure if that will work.. This also presents the same issue:

```
BOOST_AUTO_TEST_CASE(minimal_example)
{
    problem prob{zdt{1u, 30u}};
    population pop1{prob, 3u, 23u};

    moead user_algo1{100u, "grid", "tchebycheff", 1u, 1., 0.5, 20., 0.9, 2u, true, 23u};
    user_algo1.set_verbosity(1u);
    pop1 = user_algo1.evolve(pop1);
}
```

Sceki:
perhaps by putting the min pop size to 3, and the min neigh size to 2 would fix it.. I will check

bluescarni:
Ok, thanks for checking this out! Pinging @darioizzo .

Sceki:
I indeed think that enforcing `m_neighbours>1` and `popsize>=3` would fix the issue.. not sure if this can also be solved internally in the algo.. but I do not personally think it is even worthy to check inside the algo since a population size of 3 seems already small enough

Sceki:
Let me know in case you want me to do a PR !

[...]

PR #320, opened 2019-07-05
bug fix moead

Sceki:
MOEA/D presents a bug if `m_neighbours<2` ( #318 ), this PR aims to introduce a throw when this happens.

### DIFF
--- a/include/pagmo/algorithms/moead.hpp
+++ b/include/pagmo/algorithms/moead.hpp
@@ -97,7 +97,7 @@ public:
      * and Qingfu Zhang paper
      * @param seed seed used by the internal random number generator (default is random)
      * @throws value_error if gen is negative, weight_generation is not one of the allowed types, realb,cr or f are not
-     * in [1.0] or m_eta is < 0
+     * in [1.0] or m_eta is < 0, if neighbours is <2
      */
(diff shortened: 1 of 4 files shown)


Sceki:
#318

bluescarni:
@Sceki  would you mind adding also as a test case the code that resulted in an infinite loop previously? Don't have to ``BOOST_CHECK()`` anything, just put the offending code in a new ``BOOST_AUTO_TEST_CASE``.

Sceki:
@bluescarni  The problem is that that situation is never gonna happen now. If I try to reproduce it the algorithm is going to throw (because `m_neighbours=1`)

bluescarni:
> @bluescarni The problem is that that situation is never gonna happen now. If I try to reproduce it the algorithm is going to throw (because `m_neighbours=1`).
> 
> That is why I only added :
> 
> ```
>     BOOST_CHECK_THROW((moead{10u, "grid", "tchebycheff", 1u, 1., 0.5, 20., 0.9, 2u, true, 23u}), std::invalid_argument);
> ```
> 
> in the tests

You're right, sorry about the brainfart :)

@darioizzo good to go?

darioizzo:
good to go.

Issue #322, opened 2019-07-18
throw bug in PSO


Sceki:
In src/pso.cpp, the lines:
```
    if (m_omega < 0. || m_omega > 1.) {
        // variants using Inertia weight
        pagmo_throw(std::invalid_argument,
                    "The particles' inertia (or the constriction factor) must be in the [0,1] range, while a value of "
                        + std::to_string(m_variant) + " was detected");
    }
```

Should be :
```
    if (m_omega < 0. || m_omega > 1.) {
        // variants using Inertia weight
        pagmo_throw(std::invalid_argument,
                    "The particles' inertia (or the constriction factor) must be in the [0,1] range, while a value of "
                        + std::to_string(m_omega) + " was detected");
    }
```

`m_omega`, instead of `m_variant` should thus be used in the throw message.

Sceki:
I am doing a PR for this..

darioizzo:
:) thanks!

bluescarni:
Fixed by #323.

PR #323, opened 2019-07-18
PSO throw bug fix

Sceki:
Fixing #322


bluescarni:
@Sceki thanks for spotting this!

PR #326, opened 2019-07-23
MHACO: Multi-objective Hypervolume-based Ant Colony Optimizer

Sceki:
This PR aims to introduce a new multi-objective optimizer in PaGMO: `maco`. 
The main feature of this algorithm is to evolve the population by making use of the hypervolume computation.


Sceki:
@bluescarni If you prefer you can also leave only the latest commit (`clang-formatting`) for the tests.. I forgot a couple of things and I thus committed three times.

darioizzo:
@Sceki  Is this to be merged? Or we want to work on it on this PR? I would merge it and in case open a separate PR.

Sceki:
Yes it is.. Sure, we can work on possible improvements, in case we have some, separately

PR #327, opened 2019-07-23
add install instructions for pygmo on archlinux

acxz:
Got around to creating an Arch Linux package for pygmo. Note: It is only for Python3.
Since python2 maintenance will be dropped in ~5 months, I didn't create a python2 package for it.

### DIFF
--- a/doc/sphinx/install.rst
+++ b/doc/sphinx/install.rst
@@ -215,3 +215,19 @@ When done, type (in your build directory):
    make install
 
 To check that all went well fire-up your Python console and try the example in :ref:`quick-start example <getting_started_py>`.
+
+Installation on Arch Linux
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Pygmo is available on the `Arch User Repository
+<https://aur.archlinux.org>`__ (AUR) in Arch Linux. It is
+recommended to use an AUR helper like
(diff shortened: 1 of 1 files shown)


darioizzo:
Nice thanks! Any reason ipopt, nlopt and eigen are not listed as dependencies? In other words: PAGMO_WITH_EIGEN3,  PAGMO_WITH_IPOPT, PAGMO_WITH_NLOPT are activated when installing the pagmo headers?

acxz:
The reason why eigen, ipopt, and nlopt are not listed as a dependencies for pygmo is because they are not listed as dependencies on the documentation for pygmo. Pagmo is a dependency for pygmo which is listed, and if pagmo is built with said dependencies than pygmo will also have them. 

    On Wednesday, July 24, 2019, 5:34:47 AM EDT, Dario Izzo <notifications@github.com> wrote:  
 
 
Nice thanks! Any reason ipopt, nlopt and eigen are not listed as dependencies? In other words: PAGMO_WITH_EIGEN3, PAGMO_WITH_IPOPT, PAGMO_WITH_NLOPT are activated when installing the pagmo headers?

darioizzo:
ok thanks!! I had deleted my comment as I relaized the answer .... :)

PR #347, opened 2019-09-26
More FP comparison fixes.

bluescarni:


bluescarni:
@darioizzo can you take a look?

darioizzo:
sounds good.

PR #348, opened 2019-09-27
Introducing the BFE in the PSO GEN algorithm

Sceki:
This PR aims to introduce the batch fitness evaluation (i.e., `bfe`) scheme in the `pso_gen` algorithm.

### DIFF
--- a/src/algorithms/pso_gen.cpp
+++ b/src/algorithms/pso_gen.cpp
@@ -387,7 +387,7 @@ population pso_gen::evolve(population pop) const
                 best_neighb = X[0];
 
                 for (decltype(swarm_size) p = 1; p < swarm_size; p++) {
-                    if (fit[p] < best_fit) {
+                    if (detail::less_than_f(fit[p][0], best_fit[0])) {
                         best_fit = fit[p];
                         best_neighb = X[p];
                     }
@@ -406,7 +406,7 @@ population pso_gen::evolve(population pop) const
                 best_neighb = X[0];
 
                 for (decltype(swarm_size) p = 1; p < swarm_size; p++) {
-                    if (fit[p] < best_fit) {
+                    if (detail::less_than_f(fit[p][0], best_fit[0])) {
                         best_fit = fit[p];
                         best_neighb = X[p];
                     }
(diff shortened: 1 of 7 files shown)


src/algorithms/pso_gen.cpp
```diff
@@ -362,26 +364,86 @@ population pso_gen::evolve(population pop) const
...
-                fit[p] = prob.fitness(X[p]);
+            if (m_bfe) {
+                // bfe is available:
+                vector_double decision_vectors(swarm_size * dim);
+
+                decltype(decision_vectors.size()) pos = 0u;
+
+                for (decltype(swarm_size) i = 0; i < swarm_size; ++i) {
+                    // I store the individuals in a contiguous vector
+                    for (decltype(X[i].size()) ii = 0u; ii < X[i].size(); ++ii) {
```
darioizzo: same as above

src/algorithms/pso_gen.cpp
```diff
@@ -362,26 +364,86 @@ population pso_gen::evolve(population pop) const
...
-            best_neighb = X[0];
+                decltype(decision_vectors.size()) pos = 0u;
 
-            for (decltype(swarm_size) p = 1; p < swarm_size; p++) {
-                if (fit[p] < best_fit) {
-                    best_fit = fit[p];
-                    best_neighb = X[p];
+                for (decltype(swarm_size) i = 0; i < swarm_size; ++i) {
+                    // I store the individuals in a contiguous vector
+                    for (decltype(X[i].size()) ii = 0u; ii < X[i].size(); ++ii) {
```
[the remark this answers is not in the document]
Sceki: Done!
Thanks for the suggestion.

src/algorithms/pso_gen.cpp
Sceki: done as well.

src/algorithms/pso_gen.cpp
```diff
@@ -362,26 +364,76 @@ population pso_gen::evolve(population pop) const
...
+                    for (decltype(fit[i].size()) i_fit = 0; i_fit < fit[i].size(); ++i_fit) {
+                        fit[i][i_fit] = fitnesses[i];
+                        lbfit[i][i_fit] = fitnesses_lb[i];
+                    }
+                }
+                best_fit = fit[0];
+                best_neighb = X[0];
+
+                for (decltype(swarm_size) p = 1; p < swarm_size; p++) {
+                    if (fit[p] < best_fit) {
```
bluescarni: @darioizzo these fitness comparisons should be done using ``less_than_f`` right?

darioizzo: Yeps, fprgot about it .... its a pattern that is to be fixed in many other UDAs. @Sceki When comparing a fitness value to another using the ```<``` operator, the result is basically subject to issues in cases where the fitness is ```nan``` (happes if the UDP ihas numerical instabilities and returns nan). In C++ any comparison agains nan fails. So ```0.3 < nan``` is false and you loose a good individual.

To avoid this, we have  in ```detail\custom_comparisons.hpp``` a helper function called ```less_than_f ``` that places ```nans``` after ```inf```. so that ```3<inf<nan```. Can you use it here throughout?

btw also in gaco and any other UDA you happen to work on .....

Sceki: Ok, I will do it!

I think I have used it for the algos I implemented as we already discussed this in the past..

bluescarni:
@Sceki merged, thanks!

PR #349, opened 2019-09-28
bug fix for PSO GEN

Sceki:
This PR aims to fix a bug in the `pso_gen` algorithm.
Instead of
```
                        for (decltype(x1.size()) k = 0u; k < x1.size(); ++k) {
                            if (ub[j] > lb[j]) {
                                acc += (x1[k] - x2[k]) * (x1[k] - x2[k]) / (ub[k] - lb[k]) / (ub[k] - lb[k]);
                            } // else 0
                        }
```
it should be :
```
                        for (decltype(x1.size()) k = 0u; k < x1.size(); ++k) {
                            if (ub[k] > lb[k]) {
                                acc += (x1[k] - x2[k]) * (x1[k] - x2[k]) / (ub[k] - lb[k]) / (ub[k] - lb[k]);
                            } // else 0
                        }
```
As `j` spans through the `pop_size`, whereas `k` spans through the `dvs` size (which is the length of `ub`, `lb`).. This error was not spotted in the current tests because the `pop_size` were always chosen to be smaller than the `dvs` size.. thus shadowing the error.

### DIFF
--- a/src/algorithms/pso_gen.cpp
+++ b/src/algorithms/pso_gen.cpp
@@ -441,7 +441,7 @@ population pso_gen::evolve(population pop) const
                         auto x2 = X[j];
                         double acc = 0.;
                         for (decltype(x1.size()) k = 0u; k < x1.size(); ++k) {
-                            if (ub[j] > lb[j]) {
+                            if (ub[k] > lb[k]) {
                                 acc += (x1[k] - x2[k]) * (x1[k] - x2[k]) / (ub[k] - lb[k]) / (ub[k] - lb[k]);
                             } // else 0
                         }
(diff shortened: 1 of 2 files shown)


darioizzo:
A bug indeed. Nice catch. We tried with @bluescarni to understand when and how it was introduced, its likely it was me that when coding pso_gen decided to check the bounds in a separate for loop and mistyped the indexes. The issue was not caught as *pso_gen* has been rarely used since without bfe *pso* is almost always a better option. The reason *pso_gen* was coded at all in pagmo2 is to allow neuroevolution, since we used this algorithm during the ACT-MIT SPHERES experiments.

Its a definite merge for me.

PR #353, opened 2019-10-12
[WIP] remove check for python3 component of boostfor boost>v1.7.1

acxz:
In the latest Boost version, they have changed how python components need to specified. Specifically, there is no need to specify major or minor version of python, Boost takes care of it internally. In fact, if you do specify a version that CMake will error out. As of now this is causing the `python-pygmo` package for ArchLinux to break.
Seems like there is another issue that I need to tackle before this PR is ready.


acxz:
NVM, I was misguided on the fix. See: #354

PR #354, opened 2019-10-13
Fix call to find Boost components via list

acxz:
In the latest version of Boost (1.71), the quotations around a list variable is not a valid parameter for the components. Removing the quotations, allows for CMake to successfully finish.


bluescarni:
@acxz thanks for the PR! Looks good to me. Do you need a new point release for this or are you going to patch the arch linux build scripts?

acxz:
On the off chance others are using the latest Boost release, I think it would be important to have another point release.

Issue #382, opened 2020-01-06
Segfault during evolution with MHACO


ow97:
## Overview
During evolution of archipelago with MHACO a segfault occurs

## Valgrind Stack Trace
```
==1899076== Thread 20:
==1899076== Invalid read of size 8
==1899076==    at 0xBBFD28F: pagmo::hv3d::contributions(std::vector<std::vector<double, std::allocator<double> >, std::allocator<std::vector<double, std::allocator<double> > > >&, std::vector<double, std::allocator<double> > const&) const (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xBBF64AC: pagmo::hypervolume::contributions(std::vector<double, std::allocator<double> > const&) const (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xBBEFCC8: pagmo::maco::evolve(pagmo::population) const (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xB545FE0: pagmo::detail::algo_inner<pagmo::maco>::evolve(pagmo::population const&) const (in /home/myname/path/to/project/svglo/build/Release/svglo.node)
==1899076==    by 0xBA41091: pagmo::algorithm::evolve(pagmo::population const&) const (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xBC0F919: pagmo::thread_island::run_evolve(pagmo::island&) const (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xBA569BD: pagmo::island::evolve(unsigned int)::{lambda()#1}::operator()() const [clone .isra.0] (in /usr/lib/libpagmo.so.2.0)
==1899076==    by 0xBA58573: std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::__future_base::_Task_state<pagmo::island::evolve(unsigned int)::{lambda()#1}, std::allocator<int>, void ()>::_M_run()::{lambda()#1}, void> >::_M_invoke(std::_Any_data const&) (in /usr/lib/libpagmo.so.2.0)
[...]

PR #383, opened 2020-01-08
Fix for #382 (nan in hv computations)

darioizzo:
Should fix the problem of hv algorithms when nans are present (and m_verify is true)

### DIFF
--- a/src/utils/hypervolume.cpp
+++ b/src/utils/hypervolume.cpp
@@ -421,6 +421,13 @@ void hypervolume::verify_after_construct() const
             pagmo_throw(std::invalid_argument, "All point set dimensions must be equal.");
         }
     }
+    for (const auto &point : m_points) {
+        for (auto value : point) {
+            if (std::isnan(value)) {
+                pagmo_throw(std::invalid_argument, "A nan value has been encountered in the hypervolume points. Cannot construct the hypervolume object");
+            }
+        }
+    }
 }
 
 // Verify before compute method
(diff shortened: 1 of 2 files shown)


bluescarni:
There are two verification functions, ``verify_after_construct()`` and ``verify_before_compute()``. Do they both need the NaN check or is it sufficient in only one of them?

darioizzo:
Yes, it some crazy overengineered verification code. Only the check after construction is needed in this case as that happens first and always (if m_verify is true).

bluescarni:
Looks good to me, I'd like to wait for @ow97 to test the code before merging though, if possible.

bluescarni:
Windows is complaining about the test code:

```
C:\projects\pagmo2\tests\hypervolume.cpp(219): error C2124: divide or mod by zero [C:\projects\pagmo2\build\tests\hypervolume.vcxproj]
```

Probably better to replace it with a ``quiet_NaN()`` I guess?

ow97:
> Looks good to me, I'd like to wait for @ow97 to test the code before merging though, if possible.

I'll get back to you ASAP

ow97:
I'm happy that throws an exception rather than segfaulting :+1:

bluescarni:
> I'm happy that throws an exception rather than segfaulting +1

Just to be sure I understood correctly, does this mean that you tried the patch and it works fine for you?

ow97:
> > I'm happy that throws an exception rather than segfaulting +1
> 
> Just to be sure I understood correctly, does this mean that you tried the patch and it works fine for you?

Yes it does. I am in favour of merging this PR :+1:

bluescarni:
> Yes it does. I am in favour of merging this PR +1

Ok, thanks for the report and the testing! We'll probably do a 2.13 release within the week.

PR #387, opened 2020-01-27
Pygmo removal

bluescarni:


darioizzo:
Mi sembra vada bene, ho solo visto che la parte python is still present in the appveyor files ... are we still needing that CI in pagmo?

bluescarni:
> Mi sembra vada bene, ho solo visto che la parte python is still present in the appveyor files ... are we still needing that CI in pagmo?

I did not bother trying to remove the Python bits because the appveyor script is still being used to test MinGW builds. I think that's the only remaining trace of Python.

PR #392, opened 2020-02-11
Fail fast Pareto dominance

ow97:
## Description
- If we find a dimension which precludes dominance, then return `false` early
-----
## Results
Failing fast in this way yielded a ~30% speedup of this method in my program. `perf` data:
- Previous - 20.0% samples had `pareto_dominance` in the call stack
- Modified - 13.5% samples had `pareto_dominance` in the call stack
----
## Possible issues
- More performance could still possibly be squeezed out by re-arranging the order in which the comparisons are performed inside the loop
- I haven't had chance to test on a diverse range of problems


bluescarni:
@ow97 thanks for the patch! Looks good to me, but I'll ask @darioizzo to take a look as well before merging.

PR #395, opened 2020-02-14
Binder

mlooz:
Added config files for binder support


darioizzo:
cool! thanks ... the best is to create a pagmo2-binder repo under the ESA org and have it there. Here in pamgo2 we just leave a badge with the link.

PR #397, opened 2020-02-17
Implementation of xling support for repr of type erased classes

darioizzo:


bluescarni:
* I would rename ``support_cling.hpp`` to ``support_xeus_cling.hpp`` (including the ``#include`` guard) and the macro to ``PAGMO_IMPLEMENT_XEUS_CLING_REPR``, as this machinery is used by ``xeus-cling`` (the cling jupyter backend), but not by ``cling`` itself (i.e., the command-line cling prompt won't use any of this).
* When invoking ``PAGMO_IMPLEMENT_XEUS_CLING_REPR``, I would drop the ``pagmo::`` specification from the class name (i.e., ``PAGMO_IMPLEMENT_XEUS_CLING_REPR(algorithm)`` rather than ``PAGMO_IMPLEMENT_XEUS_CLING_REPR(pagmo::algorithm)``), as it is not needed.

darioizzo:
BOOST_CLASS_TRACKING(pagmo::algorithm, boost::serialization::track_never)
 
also here you would remove the ```pagmo::``` ?

bluescarni:
> BOOST_CLASS_TRACKING(pagmo::algorithm, boost::serialization::track_never)
> 
> also here you would remove the `pagmo::` ?

no, that macro adds stuff in the ``boost::`` namespace, thus we have to fully qualify ``pagmo::`` when referring to our classes. The cling macro adds stuff to the ``pagmo::`` namespace, so we don't need the namespace when we refer to pagmo classes.

PR #398, opened 2020-02-18
Docs

mlooz:
Adapted Tutorials _Using Problem_, _Using Algorithm_ and _Using Populations_ from Pygmo to C++


darioizzo:
Cool stuff thanks for the PR!
I personally love the idea of having the binder button on the tutorials in sphinx.

As for the PR, I think its better not to mimic the pygmo tutorials as we are actually running away from that logic .... I think the best is to wait for @bluescarni to define a new logic for the c++ turorial sequence. Then a PR is welcome to add to each tutorial the corresponding binder button.

Maybe you can start with the first few that are there already (pinging @bluescarni who started these c++ tutorial overhaul)

mlooz:
Thanks for the comments! That makes sense, I'll focus on making the existing ones interactive instead.

PR #402, opened 2020-02-27
Trivial: Fixed error message in CEC2009 test suite

mlooz:


bluescarni:
@mlooz thanks a lot! The failures are due to Ubuntu Cosmic being deprecated in our CI provider, I have a fix that I will apply in another PR.

PR #405, opened 2020-03-05
Topology enhancements

bluescarni:


bluescarni:
Includes so far:

- topology conversion to BGL format,
- the free form topology (aka custom topology from pagmo 1).

PR #410, opened 2020-03-24
Add runtime type extraction to the UDx

bluescarni:


doc/sphinx/install.rst
```diff
@@ -60,6 +60,7 @@ to the channels, and then we can immediately install pagmo:
 .. code-block:: console
 
    $ conda config --add channels conda-forge
+   $ conda config --set channel_priority strict
```
darioizzo: out of curiosity why was this needed?

include/pagmo/algorithm.hpp
```diff
@@ -640,6 +668,58 @@ class PAGMO_DLL_PUBLIC algorithm
...
+     * to the internal UDA instance. Differently from
+     * :cpp:func:`~pagmo::algorithm::extract()`, this function
+     * does not require to pass the correct type
+     * in input. It is however the user's responsibility
+     * to cast the returned void pointer to the correct type.
+     *
+     * .. note::
+     *
+     *    The returned value is a raw non-owning pointer: the lifetime of the pointee is tied to the lifetime
+     *    of ``this``, and ``delete`` must never be called on the pointer.
```
darioizzo: What happens if delete IS called? Can you just add a brief explanation?

include/pagmo/rng.hpp
```diff
@@ -57,7 +57,11 @@ using random_engine_type = std::mt19937;
  */
 struct PAGMO_DLL_PUBLIC random_device {
     static unsigned next();
-    static void set_seed(unsigned);
+
+#if !defined(PAGMO_DOXYGEN_INVOKED)
```
darioizzo: Was this the cause of doxygen troubles??

src/detail/type_name.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace pagmo
+{
+
+namespace detail
+{
+
+std::string demangle_from_typeid(const char *s)
+{
+#if defined(__GNUC__) || (defined(__clang__) && !defined(_MSC_VER))
```
darioizzo: e che cazzo! ma mettersi daccordo no? in questo caso mi pare che MSVC faccia la cosa piu sana :)

doc/sphinx/install.rst
bluescarni: This was added recently to the conda-forge instructions on their website.

include/pagmo/algorithm.hpp
bluescarni: Undefined behaviour of course :) This is just the standard warning that is there on all ``extract()`` functions.

include/pagmo/rng.hpp
bluescarni: No idea, this is just the workaround for the current doxygen version (1.8.17). Other doxygen versions give errors in other places.

Issue #421, opened 2020-04-22
bug in SBX implementation 


MLopez-Ibanez:
https://github.com/esa/pagmo2/blob/ab1f3a945a4b78886bb270532d80215f08782a3a/src/algorithms/nsga2.cpp#L439-L442

site1 and site2 above will be in `[0, Di-1]`, however, it should be in `[Dc, D]` as the chromosome positions are `[0, ..., Dc, Dc+1, ...., Dc+Di]`. 

Also, there are two nested loops and each inner may overwrite the whole children. That doesn't seem correct. I think it should be:

```c++
    // This implements two-point binary crossover and applies it to the integer part of the chromosome
    if (Di == 0)
        return;

    if (drng(m_e) < m_cr) {
        std::uniform_int_distribution<decltype(Dc)> ra_num(Dc, D - 1);
        auto site1 = ra_num(m_e);
        auto site2 = ra_num(m_e);
        if (site1 > site2) {
            std::swap(site1, site2);
        }
        auto j = Dc;
        for (; j < site1; ++j) {
            child1[j] = parent1[j];
            child2[j] = parent2[j];
        }
        for (; j < site2; ++j) {
            child1[j] = parent2[j];
            child2[j] = parent1[j];
        }
        for (; j < D; ++j) {
            child1[j] = parent1[j];
            child2[j] = parent2[j];
        }
    } else {
        for (auto i = Dc; i < D; ++i) {
            child1[i] = parent1[i];
            child2[i] = parent2[i];
        }
    }
```

MLopez-Ibanez:
Another minor bug is that:
```c++
if (drng(m_e) <= m_cr)
```
is called two times, hence, if we have a mixed-integer problem, this means that the crossover rate per solution is doubled (there are two chances per crossover call that the solution will be modified). 

I think the code should do once and at the very start:
```c++
if (drng(m_e) >= m_cr) return
```

MLopez-Ibanez:
Actually, since child1 is a copy of parent1 and child2 is a copy of parent2, then we can simply do:
```c++
    // This implements two-point binary crossover and applies it to the integer part of the chromosome
    if (Di == 0)
        return;

    if (drng(m_e) < m_cr) {
        std::uniform_int_distribution<decltype(Dc)> ra_num(Dc, D - 1);
        auto site1 = ra_num(m_e);
        auto site2 = ra_num(m_e);
        if (site1 > site2) {
            std::swap(site1, site2);
        }
[...]

Issue #420, opened 2020-04-22
Biased random coin tosses


MLopez-Ibanez:
In many places, the code implements a random coin toss as:

```c++
if (drng(m_e) <= p) // Do something
```
However, as `drng()` returns a value in `[0,1)`, using `<=` slightly biases the probability towards true. It is easy to see it in the extreme case when  `p = 0`, then we should never do something, however, the above condition may still be true. I believe the correct code should be:

```c++
if (drng(m_e) < p) // Do something
```
then with `p = 0`, we never do something and with `p=1` we always do something.

There are many examples of the above pattern: 
https://github.com/esa/pagmo2/blob/ab1f3a945a4b78886bb270532d80215f08782a3a/src/algorithms/nsga2.cpp#L391

bluescarni:
Fixed by #422.

Issue #419, opened 2020-04-20
Duplicated implementations of SBX and polynomial mutations


MLopez-Ibanez:
The implementations of NSGA2 and SGA have their own implementations of SBX and polynomial mutations. Are these the same? If so, why the duplication?

Same goes for MOEA/D and NSGA2 regarding polynomial mutation.

bluescarni:
@MLopez-Ibanez I'll let @darioizzo reply in more detail, but the gist is that pagmo is not currently meant to be a framework for building algorithms, so we haven't given much thought so far to exposing/re-using implementation details of algorithms.

Of course, I agree that it would be better to avoid the duplication, even only from a software engineering standpoint. @darioizzo do you think we can get this done for the next release?

MLopez-Ibanez:
> Of course, I agree that it would be better to avoid the duplication, even only from a software engineering standpoint. @darioizzo do you think we can get this done for the next release?

As shown in bug #421, at least the implementation of SBX is buggy, and it is duplicated all over the place (I have only looked at the NSGA2 one, so not sure if others have the same problems).

I only found out this because I had to reimplement SBX in pure python to use it in my UDA for pygmo2.

bluescarni:
> As shown in bug #421, at least the implementation of SBX is buggy, and it is duplicated all over the place (I have only looked at the NSGA2 one, so not sure if others have the same problems).

The best course of action seems to be to factor out these operators as public functions and fix their issues in the process. But I am not familiar with this part of the code, so it would be best if @darioizzo took charge.

darioizzo:
sure, np.

darioizzo:
I would suggest to add in ``utils`` a new header/cpp file called ``genetic_operators`` where to move all said implementations.

bluescarni:
Fixed by #422.

PR #422, opened 2020-04-23
Refactoring and exposing genetic operators

darioizzo:
In response to #421 #420 #419 this PR wants to improve the user experience when reading into the code of genetic types of metaheuristics.


src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+    vector_double::size_type site1, site2;
+    // Initialize the child decision vectors
+    vector_double child1 = parent1;
+    vector_double child2 = parent2;
+    // Random distributions
+    std::uniform_real_distribution<> drng(0., 1.); // to generate a number in [0, 1)
+
+    // This implements a Simulated Binary Crossover SBX and applies it to the non integer part of the decision
+    // vector
+    if (drng(random_engine) < p_cr) {
```
MLopez-Ibanez, line 75: The code would look better if you do an early exit here as:
```c++
if (drng(random_engine) >= p_cr)
   return std::make_pair(std::move(child1), std::move(child2));
```
Excessive indentation makes it harder to read.

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+                c1 = 0.5 * ((y1 + y2) - betaq * (y2 - y1));
+                beta = 1. + (2. * (yu - y2) / (y2 - y1));
+                alpha = 2. - std::pow(beta, -(eta_c + 1.));
+                if (rand01 <= (1. / alpha)) {
+                    betaq = std::pow((rand01 * alpha), (1. / (eta_c + 1.)));
+                } else {
+                    betaq = std::pow((1. / (2. - rand01 * alpha)), (1. / (eta_c + 1.)));
+                }
+                c2 = 0.5 * ((y1 + y2) + betaq * (y2 - y1));
+                if (c1 < lb[i]) c1 = lb[i];
```
MLopez-Ibanez, line 104: Note that c++17 has std::clamp: https://en.cppreference.com/w/cpp/algorithm/clamp
which is probably optimized.

 You may wish to add your own implementation of clamp and replace it with the standard once you require C++17.

There are several places in this file where you can use clamp().

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+    vector_double child1 = parent1;
+    vector_double child2 = parent2;
+    // Random distributions
+    std::uniform_real_distribution<> drng(0., 1.); // to generate a number in [0, 1)
+
+    // This implements a Simulated Binary Crossover SBX and applies it to the non integer part of the decision
+    // vector
+    if (drng(random_engine) < p_cr) {
+        for (decltype(dim_c) i = 0u; i < dim_c; i++) {
+            if ((drng(random_engine) <= 0.5) && (std::abs(parent1[i] - parent2[i])) > 1e-14 && lb[i] != ub[i]) {
```
MLopez-Ibanez, line 77: < 0.5

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+                } else {
+                    y1 = parent2[i];
+                    y2 = parent1[i];
+                }
+                yl = lb[i];
+                yu = ub[i];
+                rand01 = drng(random_engine);
+                beta = 1. + (2. * (y1 - yl) / (y2 - y1));
+                alpha = 2. - std::pow(beta, -(eta_c + 1.));
+                if (rand01 <= (1. / alpha)) {
```
MLopez-Ibanez, line 90: < (1. / alpha)

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    betaq = std::pow((rand01 * alpha), (1. / (eta_c + 1.)));
+                } else {
+                    betaq = std::pow((1. / (2. - rand01 * alpha)), (1. / (eta_c + 1.)));
+                }
+                c2 = 0.5 * ((y1 + y2) + betaq * (y2 - y1));
+                if (c1 < lb[i]) c1 = lb[i];
+                if (c2 < lb[i]) c2 = lb[i];
+                if (c1 > ub[i]) c1 = ub[i];
+                if (c2 > ub[i]) c2 = ub[i];
+                if (drng(random_engine) <= .5) {
```
MLopez-Ibanez, line 108: < .5

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+    for (decltype(dim_c) j = 0u; j < dim_c; ++j) {
+        if (drng(random_engine) < p_m && lb[j] != ub[j]) {
+            y = child[j];
+            yl = lb[j];
+            yu = ub[j];
+            delta1 = (y - yl) / (yu - yl);
+            delta2 = (yu - y) / (yu - yl);
+            rnd = drng(random_engine);
+            mut_pow = 1. / (eta_m + 1.);
+            if (rnd <= 0.5) {
```
MLopez-Ibanez, line 162: < .5

src/utils/genetic_operators.cpp
```diff
@@ -0,0 +1,203 @@
...
+                                                 const std::vector<vector_double::size_type> &non_domination_rank,
+                                                 const std::vector<double> &crowding_d,
+                                                 detail::random_engine_type &random_engine)
+{
+    if (non_domination_rank[idx1] < non_domination_rank[idx2]) return idx1;
+    if (non_domination_rank[idx1] > non_domination_rank[idx2]) return idx2;
+    if (crowding_d[idx1] > crowding_d[idx2]) return idx1;
+    if (crowding_d[idx1] < crowding_d[idx2]) return idx2;
+    std::uniform_real_distribution<> drng(0., 1.); // to generate a number in [0, 1)
+    return ((drng(random_engine) > 0.5) ? idx1 : idx2);
```
MLopez-Ibanez, line 199: >= .5

but actually it would be clearer as `drng(random_engine) < .5`

src/utils/genetic_operators.cpp
darioizzo: I personally do not like code with multiple exit points makes it more difficult to analys. Buts thats taste.

src/utils/genetic_operators.cpp
darioizzo: Yeps. This code is legacy from third party and I would rather not touch it too much.

src/utils/genetic_operators.cpp
MLopez-Ibanez: Fair enough! :)

darioizzo:
This is good to go afaic. The next level would be to expose this to the public API, but that would require to check the input consistency at each call with some impact on the UDA performances.

bluescarni:
> This is good to go afaic. The next level would be to expose this to the public API, but that would require to check the input consistency at each call with some impact on the UDA performances.

How about writing thin wrappers that add error checking to the current functions for public use?

This way pagmo's UDA can call directly the current implementations while 3rd party code can use the safer wrappers.

.vscode/settings.json
```diff
@@ -7,4 +7,5 @@
         "doc/doxygen/xml/": true,
         "doc/sphinx/_build/": true
     },
+    "C_Cpp.errorSquiggles": "Disabled",
```
bluescarni, line 10: Please remove this.

tests/sga.cpp
```diff
@@ -113,13 +113,13 @@ BOOST_AUTO_TEST_CASE(sga_evolve_test)
     // The we test that evolution is deterministic if the
     // seed is controlled.
     std::vector<sga> udas = {
-        sga{10u, .90, 1., 0.1, 1., 2u, "exponential", "gaussian", "tournament", 32u},
-        sga{10u, .90, 1., 0.1, 1., 2u, "sbx", "gaussian", "tournament", 32u},
-        sga{10u, .90, 1., 0.1, 1., 2u, "binomial", "gaussian", "tournament", 32u},
-        sga{10u, .90, 1., 0.1, 1., 2u, "single", "gaussian", "tournament", 32u},
+        //sga{10u, .90, 1., 0.1, 1., 2u, "exponential", "gaussian", "tournament", 32u},
```
bluescarni, line 116: Are these tests supposed to remain commented out?

.vscode/settings.json
darioizzo: Are you sure? squiggles are not working in our pagmo setup where headers are searched with the ```<>``` symbol and not ```""```. Without this I get consistently error from the c++ extension I use and I need to activate it each time.

tests/sga.cpp
darioizzo: no, remnants of trials ...

bluescarni:
Looks good to me, just a couple of minor points.

darioizzo:
> How about writing thin wrappers that add error checking to the current functions for public use?
> 
> This way pagmo's UDA can call directly the current implementations while 3rd party code can use the safer wrappers.

Sure, ok. So the question is: do we want the public API to expose these methods then?

.vscode/settings.json
bluescarni: Perhaps then you need a CMake extensions that is able to pick up the extra include paths from our build system?

Or maybe it's just better to remove this file entirely. I just hate that it always picks up details form developer's custom setups.

bluescarni: Actually, go ahead, get rid of it and add ``.vscode`` to the ``.gitignore``, so that we are free to configure our environments without stepping on each other's toes.

darioizzo: ok agreed

bluescarni:
> Sure, ok. So the question is: do we want the public API to expose these methods then?

@MLopez-Ibanez was asking for it, but I don't have strong opinions about it. If you keep it private however, you should move everything (header and src) into the ``detail/`` subdirectories then.

MLopez-Ibanez:
> Sure, ok. So the question is: do we want the public API to expose these methods then?

I think it would be useful. We are implementing an algorithm using pygmo and we had to implement these operators ourselves using NumPy.

darioizzo:
Ok, I will try to do it fot 2.16 then on this same PR.

Issue #325, opened 2019-07-23
Please make an example for dummies


stabilus:
I am entering the subject as a non mathematician and non data scientist. I'm a potential user of your code who is looking to solve actual down to earth practical problems. While your code apparently beautifully includes a multitude of problems and algorithms, they are really wasted on me as I simply don't get the math, the lingo or the examples.

I therefore suggest you provide a few practical examples of how your code could be put to use, i.e:
- temperature forecasts
- stock price prediction based on time series containing features like price and various Technical analysis indicators
- segmentation / classification
- computer gaming bot

just to name a few. Seeing how you implement (or even just define the world - the problem) would help tremendously.

Thanks.

P.S. I would even be willing to discuss one on one teaching and appropriate compensation.

bluescarni:
@stabilus work on improving the documentation is on the agenda, as you have seen in #285.

It would be helpful to understand what kind of tutorials you have in mind. PaGMO is a library which aims to be useful for, among others, the applications you mention, but it does require the user to do at least a little bit of coding. In particular, PaGMO is not really a modelling language in the same fashion as something like, e.g., Pyomo or AMPL, meaning that we don't provide higher-level constructs to translate a human-readable description of an optimisation problem into the lower-level code.

With that said, perhaps I have misunderstood and you have something else in mind (hence seeing some concrete examples of what you'd like to see in PaGMO would help).

stabilus:
My mistake. I am interested in PyGMO tutorials - not PaGMO. But since the two are related by family, I assumed both were managed in parallel. Apologies!

ckaldemeyer:
> I therefore suggest you provide a few practical examples of how your code could be put to use, i.e:
> 
>     * temperature forecasts
> 
>     * stock price prediction based on time series containing features like price and various Technical analysis indicators
> 
>     * segmentation / classification
> 
>     * computer gaming bot

Correct me if am not right but PaGMO/PyGMO is a framework primarily for optimization and not for forecasting, clustering, autonomous system even if optimization methods can be useful within these domains e.g. for minimizing/maximizing some measure within cluster methods.

For forecasting/prediction/clustering I would rather look into specific libraries such as [statsmodels](https://www.statsmodels.org/stable/index.html) or [tsam](https://github.com/FZJ-IEK3-VSA/tsam).

[...]

Issue #285, opened 2019-05-09
Docs improvements (feedback needed!)


bluescarni:
Our documentation website at https://esa.github.io/pagmo2/ needs some improvements, as recently pointed out in issue reports (#258, #263, #279, etc.) and general chatter on the gitter channel.

The API is fully documented, but we are a bit lacking in tutorials and introductory documentation (especially on the C++ side). The installation docs could also be improved and they are currently outdated (e.g., pagmo now a compiled library, we are lagging behind in the production of pip packages of windows so the documentation is not correct - a decision needs to made there, etc.).

I would encourage the community to list in this report ideas/criticism/complaints about our documentation, so we can bundle up the improvements into a dedicated documentation PR and hopefully improve the situation.

I'll start:

- [ ] the C++ API documentation should say which header file I should include to use a specific class, function, etc. I started doing this recently, but this information needs to be added everywhere. This will probably require some reorganization of the existing docs so that entities that live in the same header file are grouped together in the sphinx pages.

bluescarni:
Quoting @ShavaraniMahdi from #258 (regarding installation in Windows)

> I was trying to install Pagmo according to the instructions given in the documentation. I encountered some problems that took me a long time to resolve and I thought it is a good idea to share my experience in order to enrich the documentation for future references.

mpowelson:
First of all, let me say that this looks great. I have also not looked at the Python tutorials very much since I am using Pagmo via C++. However, it strikes me that most people that come across Pagmo (myself included) find it because they have a specific problem they are looking to solve. To me it seems that it would be useful to expand the C++ tutorials section to step through creating a UDP, solving it, and describing some of the knobs that we can turn to effect the results. It looks like much of this might already be in the Python tutorials.

Additionally, an overview of evolutionary algorithm terminology and concepts (or pointers to some good reference material) would be useful. For example, when we instantiate an archipelago with 16 islands each with 20 individuals, what does that mean? What can we expect increasing one of those to do to performance?

bluescarni:
@mpowelson yes, fully agreed on the first part. In fact, this sort of step-by-step guide on writing a UDP, starting with something simple and then progressing to a more complicated example, is exactly what I had in mind. We sort of have something like that in Python, but not in a well organised fashion. C++ was a bit left behind in terms of tutorials because of manpower issues, in the belief that a larger fraction of users would benefit from Python docs.

[...]

PR #431, opened 2020-05-11
Doc improvement - Extending and restructuring of the doc tutorials 

juliusroeder:
In response to #285 and #325 this PR aims to clarify and simplify the tutorials for beginners. 
It extends the initial "Writing your first optimisation problem" and merges the "Evolving a population" from the "Basics" section to give one coherent example of how to construct a UDP and solve it. 
Let me know what you think.


bluescarni:
@juliusroeder thanks a lot, this is an extremely welcome initiative!

I like the new tutorials, how they are organized and the writing style. I have a minor question about some of the links you added, e.g., ``.. _cpp_archipelago:`` on top of ``archipelago.rst``. Is there a specific reason why you added a new reference instead of just referencing directly the archipelago class via `` :cpp:class:`~pagmo::archipelago` `` (or similar)?

Another minor comment is that perhaps you could update the comment ``// Our simple example problem, version 1.`` in the new tutorials, so that the problem version is updated to 2, 3, etc.

We'll take a deeper look once you feel like the PR is ready, but I am really happy about the direction where this is going.

darioizzo:
obsolete.

PR #432, opened 2020-05-11
Improve thread re-use with task_queue pool

ow97:


ow97:
Note that the `fork_island` has been disabled on this branch. This will need fixing before merging.

ow97:
My program, when compiled against this branch, shows the new expected behaviour of O(1) threads rather than O(n). Overall runtime improves by a factor of 2-3. However there appears to be further room for improvement as only 60% CPU utilisation is achieved overall. Likely due to time spent waiting for mutexes. Does TBB provide a threadsafe and lock free queue implementation which would allow us to `enqueue` without a mutex?

bluescarni:
Hey @ow97 thanks for looking into this!

I think there are two ways in which we can proceed to reduce thread creation and destruction overhead in ``task_queue``.

The first one is essentially what you implemented here, a thread pool pattern. Implementing it on top of TBB would be the "right thing" to do I believe, because TBB gets you 99% there already with the primitives it provides, and it integrates very well with other TBB usages across pagmo (e.g., in terms of nested parallelism with thread_island, the thread-based batch fitness evaluator, etc.).

The second option, which is the one I was thinking about, would be simply to avoid destroying the thread associated to an island on island destruction, store it in a global list, and re-use it the next time a new island is created. This is semantically different from the thread pool option, because it allows the simultaneous existence of an arbitrary number of threads (whereas the thread pool pattern caps the number of threads to some limit).

Going for a thread pool pattern also has consequences that are difficult to understand immediately, because for example it changes how migration operates.

For these reasons, I'd be more oriented towards an approach in which we recycle existing thread instead of a thread pool.

ow97:
@bluescarni I'm happy to have a go at implementing this, though it sounds like you have a better idea of exactly what TBB provides than I do (I'm more familiar with OpenMP).

bluescarni:
> @bluescarni I'm happy to have a go at implementing this, though it sounds like you have a better idea of exactly what TBB provides than I do (I'm more familiar with OpenMP).

Great, thanks!

Basically TBB already has a built-in thread pool to which you can submit tasks. The way we normally use it in pagmo is via a ``task_group`` object, but I don't think that would work in ``task_queue``. Basically what we need there is to submit a task to the TBB runtime and get a ``std::future`` in return, on which we can wait. I'll do some reading.

ow97:
Fantastic, thanks for your time on this. I look forward to hearing back.

ow97:
I've had an attempt at replacing `task_queue` with TBB's `task_group`.

Performance of unit tests has taken a severe hit. Currently investigating why.

Any input welcome :)

ow97:
A lot of time is now spent in the `archipelago_bfe_ctors` (2m 4s) and `archipelago_mo_migration_bug` (2m 46s) tests. Perf shows that 39.8% of execution time of the `archipelago` test suite is spent in `libpagmo.so.6.0`pagmo::select_best_N_mo` and 35.6% of time spent in `libpagmo.so.6.0`pagmo::population::push_back`. Not sure why this seems to have changed so drastically.

ow97:
Spent the afternoon investigating, and I'm afraid to report I'm out of ideas. @bluescarni if you've got time to have a look, that would be much appreciated.

bluescarni:
> Spent the afternoon investigating, and I'm afraid to report I'm out of ideas. @bluescarni if you've got time to have a look, that would be much appreciated.

I was reading through the code, but I haven't spotted anything yet. How is the cpu usage during the tests? Are all cores saturated?

ow97:
Yeah 99.0% - 99.9% on all of 8 cores (4 hyper-threaded)

bluescarni:
> A lot of time is now spent in the `archipelago_bfe_ctors` (2m 4s) and `archipelago_mo_migration_bug` (2m 46s) tests. Perf shows that 39.8% of execution time of the `archipelago` test suite is spent in `libpagmo.so.6.0 pagmo::select_best_N_mo` and 35.6% of time spent in `libpagmo.so.6.0 pagmo::population::push_back`. Not sure why this seems to have changed so drastically.

``pagmo::select_best_N_mo`` is an algorithm with a rather heavy complexity, ``O(MN^2)`` (where N is the number of individuals). I am wondering if somehow the use of the thread pool somehow changes something in the migration framework to the point that that function is called with a much larger number of individuals in input than usual.

bluescarni:
Could you perhaps modify the source code of that function to print the value of ``input_f.size()`` and ``N`` as soon as it is invoked? And then compare with the old task_queue implementation.

ow97:
See also https://github.com/esa/pagmo2/pull/433

ow97:
As per @bluescarni's comment:
> For these reasons, I'd be more oriented towards an approach in which we recycle existing threads instead of a thread pool.
I've opened #433. Given that implementation looks just as promising as this one, I will now close this PR.

PR #435, opened 2020-05-20
Simplifications/cleanups for task_queue.

bluescarni:


bluescarni:
@ow97 how does this look to you?

The main implementation change is that now we deal directly with tasks represented as a ``std::packaged_task``, instead of going through a lambda which eventually ends up in a ``std::function``. I am not 100% sure why I didn't implement it this way since the beginning, but probably it was a case of defensive programming as I hadn't understood fully at the time what one could (not) do with ``std::packaged_task``.

Another simplification is that I have removed the ``m_stop`` check when queuing a task. This is not needed, as the queue can be in a stopping state only during destruction and we don't require to be able to add new tasks while destroying the queue (this would imply that we allow for concurrent operations from multiple threads on the same ``task_queue`` object, and we don't need that in pagmo).

ow97:
Looks good to me :+1:

PR #437, opened 2020-05-30
Updating grey wolf optimizer

ahmed2m:
There was a [problem](https://github.com/7ossam81/EvoloPy/issues/21) in the code implemented before in [EvoloPy](https://github.com/7ossam81/EvoloPy), the code which pagmo's GWO implementation was based on intially, this mistake originates from the [matlab function](https://www.mathworks.com/matlabcentral/fileexchange/44974-grey-wolf-optimizer-gwo) purposed in the original paper.  The author confirmed in mathworks comment section (in 2014) that this should be corrected.

The _actual_ position updating has no issue but the best scores/positions logged for each iteration isn't updating the three values (Alpha, Beta and Delta) it was only one of the three only when the best of current generation is better than the value the tier x holds right now.

Example: 
```
A: 1  B: 1.9  D: 2.3
Next iter Best 3 are:  0.9        1.1           2.5
A: 0.9 B: 1.1   D: 2.3
```
While the correct values should be
`A: 0.9 B: 1.1   D: 1.9`

The original GWO PR is #268. I took a look at it and learned about exposing the code to pygmo, so I compiled pagmo and installed pygmo from the source and re-generated the [comparison](https://gist.github.com/michiboo/2df766cd7f59c0eb579290c8cbf2536e) made in the original PR to show how different the two approaches are. And it shows slight noticeable difference.

current GWO:
OLDALL

After updating:
NEW_ALL COMP

I didn't change much of the code, I don't think any tests should be refactored from this change.

### DIFF
--- a/src/algorithms/gwo.cpp
+++ b/src/algorithms/gwo.cpp
@@ -167,11 +167,17 @@ population gwo::evolve(population pop) const
             auto fit = pop.get_f()[i];
             // Update alpha, beta and delta
             if (fit[0] < alpha_score) {
+                delta_score = beta_score;
+                delta_pos = beta_pos;
+                beta_score = alpha_score;
+                beta_pos = alpha_pos;
                 alpha_score = fit[0];
                 alpha_pos = agents_position[i];
             }
 
             if (fit[0] > alpha_score && fit[0] < beta_score) {
+                delta_score = beta_score;
+                delta_pos = beta_pos;
                 beta_score = fit[0];
                 beta_pos = agents_position[i];
             }


darioizzo:
Thanks for the PR.

GWO should be tested only on translated problems (i.e. problems where the optimal solution is not at the origin.

The reason is that the algorithm (as originally published) is **cheating** as it assumes implicitly that the optimal solution is in the origin in which case (see rastrigin, rosenbrock and griewank plots) it may seem as to be outperforming other algorithms, while in general its heuristic is of no use.

If you can make the CI pass and change GWO heuristic as to imporve its performances on translated problems, we could also consider merging this, otherwise our plan would be to eventually remove GWO from the official pagmo algos.

darioizzo:
Merging this, even though the GWO should be removed from pagmo eventually as its shady and seems just wrong ...

Issue #438, opened 2020-06-22
[BUG] The new implementation of SBX seems wrong


MLopez-Ibanez:
**Describe the bug**
Pull request #436 changed SBX. I believe the change is wrong and it doesn't match the SBX in https://www.iitk.ac.in/kangal/codes/nsga2/nsga2-gnuplot-v1.1.6.tar.gz

darioizzo:
And what makes you believe that?

darioizzo:
https://github.com/esa/pagmo2/blob/master/src/utils/genetic_operators.cpp#L101 here a few lines were lost in the refactoring, changing the distribution. The following is the corrected version:
```
                rand01 = drng(random_engine);
                beta = 1. + (2. * (y1 - yl) / (y2 - y1));
                alpha = 2. - std::pow(beta, -(eta_c + 1.));
                if (rand01 < (1. / alpha)) {
                    betaq = std::pow((rand01 * alpha), (1. / (eta_c + 1.)));
                } else {
                    betaq = std::pow((1. / (2. - rand01 * alpha)), (1. / (eta_c + 1.)));
                }
                c1 = 0.5 * ((y1 + y2) - betaq * (y2 - y1));
                beta = 1. + (2. * (yu - y2) / (y2 - y1));
                alpha = 2. - std::pow(beta, -(eta_c + 1.));
                if (rand01 < (1. / alpha)) {
                    betaq = std::pow((rand01 * alpha), (1. / (eta_c + 1.)));
                } else {
                    betaq = std::pow((1. / (2. - rand01 * alpha)), (1. / (eta_c + 1.)));
                }
                c2 = 0.5 * ((y1 + y2) + betaq * (y2 - y1));
```

MLopez-Ibanez:
See pr #439

MLopez-Ibanez:
Also, the changes to SBX and polynomial mutation operators are not described in the Changelog for 2.16: https://esa.github.io/pagmo2/changelog.html

Those changes seem newsworthy to me. People may get different results in their algorithms after updating.

bluescarni:
@MLopez-Ibanez perhaps you can add the entry in the changelog in #439?

MLopez-Ibanez:
Done

bluescarni:
Fixed by #439.

PR #439, opened 2020-06-22
Fix bug #438 SBX operator

MLopez-Ibanez:
@darioizzo

### DIFF
--- a/src/utils/genetic_operators.cpp
+++ b/src/utils/genetic_operators.cpp
@@ -43,6 +43,19 @@ see https://www.gnu.org/licenses/. */
 #include <pagmo/utils/generic.hpp>
 #include <pagmo/utils/genetic_operators.hpp>
 
+namespace {
+// Helper function for implementation of the binary crossover.
+double sbx_betaq(double beta, double eta_c, double rand01)
+{
+    double alpha = 2. - std::pow(beta, -(eta_c + 1.));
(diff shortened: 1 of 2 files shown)


src/utils/genetic_operators.cpp
```diff
@@ -49,6 +49,23 @@ namespace pagmo
 namespace detail
 {
 
+// Helper function for implementation of the binary crossover.
+static inline double sbx_betaq(double beta, double eta_c, detail::random_engine_type &random_engine)
```
darioizzo, line 53: please remove static inline and move to anonymous namespace.

src/utils/genetic_operators.cpp
```diff
@@ -49,6 +49,23 @@ namespace pagmo
...
+{
+    // Random distributions
+    std::uniform_real_distribution<> drng(0., 1.); // to generate a number in [0, 1)
+    
+    double alpha = 2. - std::pow(beta, -(eta_c + 1.));
+    double rand01 = drng(random_engine);
```
darioizzo, line 59: here you will be using two different drng for the two calls. In the original deb code, intead, the same is used.
maybe pass rand01 as a paramter and remove tthe rng inside here?

bluescarni:
@MLopez-Ibanez thanks a lot, merging.

PR #448, opened 2020-09-16
Fix invalid indexing

Doodle1106:
If I understand it correctly we should use K to access each element of ub and lb here.

### DIFF
--- a/src/algorithms/pso.cpp
+++ b/src/algorithms/pso.cpp
@@ -409,7 +409,7 @@ population pso::evolve(population pop) const
                         auto x2 = X[j];
                         double acc = 0.;
                         for (decltype(x1.size()) k = 0u; k < x1.size(); ++k) {
-                            if (ub[j] > lb[j]) {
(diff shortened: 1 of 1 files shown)


darioizzo:
It seems like a clear bug. Its strange though as 

1) in pso_gen https://github.com/esa/pagmo2/blob/master/src/algorithms/pso_gen.cpp#L499 the lines look correct and pso_gen is essentially cut and paste from pso

2) if I read the current code correctly it should go in seg fault when pso is called on a popsize > problem dimension ....

is 2) the case?

Doodle1106:
yes, seg fault happens when popsize is bigger than problem dimension.  I fixed my local code as described in the PR and now pso works as predicted, no seg fault.

darioizzo:
wow, thanks for catching it. I still wonder how this could happen since pso_gen is correct and is a copy of pso ... mah lets not dig further .. it likely some idiotic commit from me :)

PR #450, opened 2020-09-18
add reference for mhaco

Sceki:
PR for adding the reference to MHACO now that is available at https://ieeexplore.ieee.org/document/9185694.


bluescarni:
Congrats on the paper!

Sceki:
Grazie Francesco :)

PR #452, opened 2020-09-22
added genetic_operators.hpp to include list in pagmo.hpp

mlooz:
This causes it to be installed by CMake.
Without this, it will, at least on my Ubuntu 18.04. not end up being installed and not found by the current master of pygmo, which looks for it when the pagmo version is newer than 2.15.


bluescarni:
Thanks, looks like we forgot to add the header. I'll merge this.

I am a bit confused however what the problem is in pygmo? As far as I can see, pygmo never includes ``pagmo.hpp``, and the fact that ``genetic_operators.hpp`` is included or not in ``pagmo.hpp`` should have zero effects on the CMake side.

mlooz:
I don't know how the interaction happened, but before this change, genetic_operators.hpp was not installed by CMake.

PR #454, opened 2020-09-24
fix in unconstrain meta problem

mlooz:
NaNs in weighted constraints now lead to NaNs in fitness

### DIFF
--- a/src/problems/unconstrain.cpp
+++ b/src/problems/unconstrain.cpp
@@ -167,7 +167,7 @@ vector_double unconstrain::fitness(const vector_double &x) const
                 }
             }
             for (decltype(nc) i = 0u; i < nc; ++i) {
-                if (!(c[i] < 0.)) {
+                if (!(c[i] <= 0.)) {
                     penalty += m_weights[i] * c[i];
                 }
             }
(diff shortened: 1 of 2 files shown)


src/problems/unconstrain.cpp
```diff
@@ -167,7 +167,7 @@ vector_double unconstrain::fitness(const vector_double &x) const
                 }
             }
             for (decltype(nc) i = 0u; i < nc; ++i) {
-                if (c[i] > 0.) {
+                if (!(c[i] < 0.)) {
```
bluescarni, line 170: Shouldn't this be ``<=`` rather than ``<`` to match the original code for non-NaN values?

tests/unconstrain.cpp
```diff
@@ -32,6 +32,7 @@ see https://www.gnu.org/licenses/. */
 
 #include <boost/lexical_cast.hpp>
 #include <boost/test/tools/floating_point_comparison.hpp>
+#include <boost/math/special_functions/fpclassify.hpp>
```
bluescarni, line 35: No need to use boost for ``isnan``, just include ``<cmath>`` and use ``std::isnan()``.

src/problems/unconstrain.cpp
mlooz: True! It leads to the same behavior in practice, since it only differs for c[i] == 0, where then 0 is added to the penalty. Still, pushed a change to make it consistent.

tests/unconstrain.cpp
mlooz: Done!

Issue #445, opened 2020-08-28
[BUG] Installs cmake scripts with missing NLOPT::header and IPOPT::header targets


yurivict:
Installed cmake scripts:
```
$ grep NLOPT::header `pkg info -l pagmo2`
grep: pagmo2-2.15.0.95:: No such file or directory
/usr/local/lib/cmake/pagmo/FindNLOPT.cmake:    if(NLOPT_FOUND AND NOT TARGET NLOPT::header)
/usr/local/lib/cmake/pagmo/FindNLOPT.cmake:        message(STATUS "Creating the 'NLOPT::header' imported target.")
/usr/local/lib/cmake/pagmo/FindNLOPT.cmake:        add_library(NLOPT::header INTERFACE IMPORTED)
/usr/local/lib/cmake/pagmo/FindNLOPT.cmake:        set_target_properties(NLOPT::header PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${NLOPT_INCLUDE_DIR}")
/usr/local/lib/cmake/pagmo/pagmo_export.cmake:  INTERFACE_LINK_LIBRARIES "Threads::Threads;Boost::boost;Boost::serialization;Eigen3::eigen3;NLOPT::header;IPOPT::header"
```

only contain the definition of ```NLOPT::header``` in your private ```FindNLOPT.cmake``` that isn't called by dependent packages, yet you export this target to dependent packages through ```pagmo_export.cmake```.

Same with ```IPOPT::header```.

This breaks DART: https://github.com/dartsim/dart/issues/1473

bluescarni:
Thanks for the report!

I am a bit confused, because invoking ``find_package(pagmo)`` should invoke the ``FindNLOPT.cmake`` script, which in turn defines the ``NLOPT::header`` target. For instance, when we compile ``pygmo`` (which invokes ``find_package(pagmo)``), we don't have any issues.

Could the problem be some naming collisions here:

https://github.com/dartsim/dart/blob/e480e2f06bd8c0dbd0be6862b7e5b4cb01df6cdf/cmake/DARTFindpagmo.cmake#L9

Here ``find_package(pagmo)`` will define some NLOPT-related targets/variables, but then DART's own machinery for detecting NLOPT is also invoked. Perhaps they are clashing?

yurivict:
> I am a bit confused, because invoking find_package(pagmo) should invoke the FindNLOPT.cmake script

[...]

PR #457, opened 2020-12-12
Fix component finding

Bidski:
Fixes the bug that was discovered in https://github.com/esa/pagmo2/issues/445


bluescarni:
Thanks for the PR, merging. I'll make a sweep of the build system to figure out if we are making the same mistake elsewhere...

I am just baffled that it seems to work on other setups...

Issue #456, opened 2020-12-10
[BUG] Pagmo needs C++17 compiler features but this is not reflected when using pagmo-config.cmake


Bidski:
**Describe the bug**
`std::conjunction`, `std::disjunction`, and `std::negation` are all only available in C++17. When including pagmo from a third-party application via CMake compilation will fail unless the third-party app also sets C++17. However, pagmo-config.cmake or pagmo_export.cmake should be setting the required compiler standard so that third-party apps don't need to be concerned with this detail.

**To Reproduce**
1. In CMakeLists.txt `find_package(pagmo CONFIG)`
2. In C++ file `#include <pagmo/pagmo.hpp>`
3. Compile
```
/usr/include/pagmo/type_traits.hpp:72:26: error: ‘conjunction’ in namespace ‘std’ does not name a template type
   72 | using conjunction = std::conjunction<Args...>;
      |                          ^~~~~~~~~~~
/usr/include/pagmo/type_traits.hpp:72:21: note: ‘std::conjunction’ is only available from C++17 onwards
   72 | using conjunction = std::conjunction<Args...>;
      |                     ^~~
/usr/include/pagmo/type_traits.hpp:75:26: error: ‘disjunction’ in namespace ‘std’ does not name a template type
   75 | using disjunction = std::disjunction<Args...>;
      |                          ^~~~~~~~~~~
/usr/include/pagmo/type_traits.hpp:75:21: note: ‘std::disjunction’ is only available from C++17 onwards
   75 | using disjunction = std::disjunction<Args...>;
      |                     ^~~
/usr/include/pagmo/type_traits.hpp:78:23: error: ‘negation’ in namespace ‘std’ does not name a template type
   78 | using negation = std::negation<T>;
      |                       ^~~~~~~~
/usr/include/pagmo/type_traits.hpp:78:18: note: ‘std::negation’ is only available from C++17 onwards
   78 | using negation = std::negation<T>;
```

**Environment (please complete the following information):**
 - OS: Arch Linux
 - Installation method: pacman package manager
 - Version: 2.16

bluescarni:
@Bidski thanks for the report.

pagmo is a C++17 library, so the assumption is that packages depending on pagmo are also accepting that they need to be compiled in C++17 mode.

[...]

PR #458, opened 2020-12-15
Set C++ standard so that it propagates to users

Bidski:
See #456 

Setting the C++ standard in this way causes it to show up in `pagmo_export.cmake` as an `INTERFACE_COMPILE_FEATURE` which, in turn, will instruct anyone linking against the `pagmo` cmake target to compile using the C++17 standard (at minimum). See [cmake docs](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#high-level-meta-features-indicating-c-standard-support)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # CMake version check.
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.8)
 # NOTE: policy for using the CheckIPOSupported module:
 # https://cmake.org/cmake/help/latest/policy/CMP0069.html
 cmake_policy(SET CMP0069 NEW)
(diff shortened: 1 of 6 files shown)


bluescarni:
@Bidski that's great, thanks a lot for digging this out.

Could I ask you to make the following further additions to the PR:

- apply the same modification to the ``CMakeLists.txt`` of ``tests``, ``benchmarks`` and ``tutorials``,
- bump up the minimum required CMake version to 3.8 in the main ``CMakeLists.txt``,
- update the documentation in ``doc/sphinx/install.rst`` to reflect the fact that the minimum required CMake version is now 3.8.

If you don't have time/bandwidth for this, I'll do this myself, just let me know.

Bidski:
@bluescarni I have applied the changes you requested

bluescarni:
@Bidski great, thanks a lot! I'll merge as soon as the CI runs.

Issue #486, opened 2021-10-04
[BUG] Consecutive calls to PSO is not necessarily the same of running with multiple generations


tarcisiofischer:
According to [this discussion](https://github.com/esa/pagmo2/issues/485), all algorithms should behave the same with `gen=N` vs `N` calls with `gen=1` (Did I understand correcly?).

I tried with PSO in pygmo, and couldn't make it work (Will explain better after showing the code):

```
def pso_run_a(problem, seed):
    population = pg.population(pg.problem(problem), N_PARTICLES, seed=seed)
    algorithm = pg.algorithm(
        pg.pso(
            gen=1,
            omega=1,
            eta1=1,
            eta2=1,
            seed=seed,
            memory=True,
        )
    )
    for i in range(N_GENERATIONS):
        population = algorithm.evolve(population)
        print("Champ> ", population.champion_f[0])
```

```
def pso_run_b(problem, seed):
    population = pg.population(pg.problem(problem), N_PARTICLES, seed=seed)
    algorithm = pg.algorithm(
        pg.pso(
            gen=N_GENERATIONS,
            omega=1,
            eta1=1,
            eta2=1,
            seed=seed,
            memory=True,
        )
    )
    # algorithm.set_verbosity(1)
    population = algorithm.evolve(population)
    print("Champ> ", population.champion_f[0])
```

```
class MyProblem:
    def fitness(self, x):
        return [np.sum(x**2)]

    def get_bounds(self):
        return ([-10, -10], [10, 10])
```

Results:
`pso_run_a:  35.042803177459504`
`pso_run_b:  14.44962519803876`

------

It seems to be a bug related to the way the population evolves. AFAIU, when returning from the algorithm, PSO will return the best fit ( [important code here](https://github.com/esa/pagmo2/blob/master/src/algorithms/pso.cpp#L438-L441) ), and then, in the next run, `X` will be in relation to `lbX`. While when running the code iteratively with multiple generations (and assuming that one generation didn't perform better than the previous, then `X` and `lbX` will be different ( [important code here](https://github.com/esa/pagmo2/blob/master/src/algorithms/pso.cpp#L361-L372) )

There are several ways of solving this. The simplier, I believe, is to update keep an internal `m_X` in the `PSO` algorithm class. But please take in consideration the bigger picture here: It doesn't seem to be easy to guarantee that the feature of N consecutive calls to be the same as `gen=N` (do you agree?)... I don't know if you do care about that, but I found it important to open up to discussion.

[...]

PR #487, opened 2021-10-05
Fb fix pso memory

tarcisiofischer:
Related to #486


src/algorithms/pso.cpp
```diff
@@ -198,25 +208,31 @@ population pso::evolve(population pop) const
...
+                best_fit = pop.get_f()[pop.best_idx()];
+                break;
+            case 2:
+            default:
+                initialize_topology__lbest(neighb);
+        }
     }
     // auxiliary variables specific to the Fully Informed Particle Swarm variant
     double acceleration_coefficient = m_eta1 + m_eta2;
-    double sum_forces;
```
tarcisiofischer, line 219: Unrelated. It was just that the variable was not initialized.
It is not a bug, since it is initialized later in the code, but I find it reasonable to avoid uninitialized stuff.

include/pagmo/algorithms/pso.hpp
```diff
@@ -254,9 +267,10 @@ class PAGMO_DLL_PUBLIC pso
     unsigned m_neighb_param;
     // memory
     bool m_memory;
+    mutable std::optional<pso::memory> m_memory_data;
```
bluescarni, line 270: ``std::optional`` might be problematic because I don't think it is supported by the Boost.serialization library. If this is indeed the case, replacing it with ``boost::optional`` should be straightforward.

bluescarni (changes requested):

bluescarni:
@tarcisiofischer thanks for the PR!

It's a +1 from me on the coding side, I'll let @darioizzo review the algorithmic part.

This is still missing serialization support though. You should make the new ``memory`` struct serializable and then ensure that the new ``pso`` data member ``std::/boost::optional<memory>`` is serialised together with the other data members. Please let me know if you need help with this part.

I was also thinking that perhaps as an alternative to the ``optional`` approach, you could use ``std::unique_ptr`` instead. This way the definition of the ``memory`` struct could go in the ``pso.cpp`` file using the pimpl pattern, in order to minimise the changes to the header file.

tarcisiofischer:
> @tarcisiofischer thanks for the PR!
> 
> It's a +1 from me on the coding side, I'll let @darioizzo review the algorithmic part.

Nice thanks :)

> This is still missing serialization support though. You should make the new `memory` struct serializable and then ensure that the new `pso` data member `std::/boost::optional<memory>` is serialised together with the other data members. Please let me know if you need help with this part.

I've implemented the serialization support. It was really straightforward, but perhaps I did missed something. Please take a look!

> I was also thinking that perhaps as an alternative to the `optional` approach, you could use `std::unique_ptr` instead. This way the definition of the `memory` struct could go in the `pso.cpp` file using the pimpl pattern, in order to minimise the changes to the header file.

Hmm... Are you concerned about the `memory` struct in the `hpp` file due to possible increase of compile time? Can you confirm I should go for it and remove the `optional`? (I ask because I prefer the `optional` solution, but at the same time I understand that this is not my code hehe Just double checking that this is important enough and I should really do it!)

include/pagmo/algorithms/pso.hpp
```diff
@@ -254,9 +271,10 @@ class PAGMO_DLL_PUBLIC pso
     unsigned m_neighb_param;
     // memory
     bool m_memory;
-    // paricles' velocities
-    mutable std::vector<vector_double> m_V;
+    mutable boost::optional<pso::memory> m_memory_data;
 
+    // particles' velocities
+    mutable std::vector<vector_double> m_V;
```
tarcisiofischer, line 277: I wonder if I should move this to `memory`, as it is the only variable that now has two semantics. (1) the usual `m_V` used as an variable in the algorithm (the others are local variables) and (2) as a memory variable.

Let me know if this makes sense to you.

bluescarni (changes requested):

bluescarni:
> I've implemented the serialization support. It was really straightforward, but perhaps I did missed something. Please take a look!

It looks good. Normally adding serialisation via Boost.serialisation is rather mechanical (one just needs to make sure that all data members are indeed included in the serialisation call).

> Hmm... Are you concerned about the `memory` struct in the `hpp` file due to possible increase of compile time? Can you confirm I should go for it and remove the `optional`? (I ask because I prefer the `optional` solution, but at the same time I understand that this is not my code hehe Just double checking that this is important enough and I should really do it!)

No it's fine, it's just that at one point I went through a major refactor moving the library from header-only to compiled and since those days I have been a bit obsessed about reducing the code in the header files :)

bluescarni:
The docs build failure is due to a web link that became dead. I'll fix it later.

tarcisiofischer:
> The docs build failure is due to a web link that became dead. I'll fix it later.

I can do it, but I need some help here.
I believe you refer to this:
```
(docs/cpp/utils/genetic_operators: line   61) broken    https://www.iitk.ac.in/kangal/papers/k2012016.pdf - 404 Client Error: Not Found for url: https://www.iitk.ac.in/kangal/papers/k2012016.pdf
```

Which seems to be automatically generated from `utils/genetic_operators.cpp`.
The difficult part here is to know what to do. Link seems to be indeed broken, and I don't know if I can get it from somewhere else...

Other than that, I believe I'm done here. Let me know if anything else is necessary, and thanks again for all the support and feedbacks!

darioizzo:
@tarcisiofischer 
Sorry for this delay .... this is good to go ... would you be willing to implement the same fix for the pso_gen? If you have not the time now we can also merge this in the meantime ....

tarcisiofischer:
> @tarcisiofischer Sorry for this delay .... this is good to go ... would you be willing to implement the same fix for the pso_gen? If you have not the time now we can also merge this in the meantime ....

Don't worry about it :)
Unfortunately, I moved away from the project that was using this :( I _may_ be able to implement the same fix eventually, but now is not a good time for me, sorry. I suggest merging this as it is, I can get back to it when I have the time to do so, if that's ok.

jschueller:
it was included in #628 
@darioizzo this can be closed also

sylvesterkaczmarek (commented):
When `m_memory_data` exists, `evolve()` replaces the freshly sized `X`, `V`, fitness and topology buffers with persisted vectors without checking the current swarm size or dimension. Reusing one memory-enabled PSO on a different population can therefore index stale buffers out of bounds; reset/reject incompatible memory and add a regression that changes population size between calls.

Issue #490, opened 2021-11-30
[BUG] gaco::pheromone_computation crashes


jschueller:
the following sample makes gaco crash:
```
#include <pagmo/algorithms/gaco.hpp>
#include <pagmo/population.hpp>
#include <pagmo/problems/rosenbrock.hpp>

using namespace pagmo;

int main()
{
        problem prob{rosenbrock{2u}};
        population pop1{prob, 1u, 23u};
        gaco user_algo1{3u, 1u, 1.0, 1e9, 0.01, 0, 7u, 1000u, 1000u, 0.0, false, 23u};
        pop1 = user_algo1.evolve(pop1);
}
```

under gdb:
```
   Gen:        Fevals:          Best:        Kernel:        Oracle:            dx:            dp:
      1              0          76143              1          1e+09              0              0
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_vector.h:1042: std::vector::reference std::vector<std::vector<double, std::allocator<double> >, std::allocator<std::vector<double, std::allocator<double> > > >::operator[](std::vector::size_type) [_Tp = std::vector<double, std::allocator<double> >, _Alloc = std::allocator<std::vector<double, std::allocator<double> > >]: Assertion '__builtin_expect(__n < this->size(), true)' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff671f859 in __GI_abort () at abort.c:79
#2  0x00007ffff7004592 in pagmo::gaco::pheromone_computation (this=<optimized out>, gen=<optimized out>, prob_cumulative=<error reading variable: Cannot access memory at address 0x8>, omega_vec=..., 
    sigma_vec=std::vector of length 2, capacity 2 = {...}, popul=..., sol_archive=...) at /home/schueller/projects/pagmo2/src/algorithms/gaco.cpp:784
#3  0x00007ffff6ffff98 in pagmo::gaco::evolve (this=<optimized out>, pop=...) at /home/schueller/projects/pagmo2/src/algorithms/gaco.cpp:273
[...]

PR #491, opened 2021-11-30
gaco bugfix: throw if ker < 2

Sceki:
Bugfix for [issue #490](https://github.com/esa/pagmo2/issues/490): `gaco` should throw if `ker` size is < 2

### DIFF
--- a/include/pagmo/algorithms/gaco.hpp
+++ b/include/pagmo/algorithms/gaco.hpp
@@ -115,7 +115,7 @@ public:
      *
      * @throws std::invalid_argument if \p acc is not \f$ >=0 \f$, \p impstop is not a
      * positive integer, \p evalstop is not a positive integer, \p focus is not \f$ >=0 \f$,
-     * \p ker is not a positive integer, \p oracle is not positive, \p
+     * \p ker is not a positive integer \in [2, pop_size], \p oracle is not positive, \p
      * threshold is not \f$ \in [1,gen] \f$ when \f$memory=false\f$ and  \f$gen!=0\f$, \p threshold is not \f$ >=1 \f$
      * when \f$memory=true\f$ and \f$gen!=0\f$, \p q is not \f$ >=0 \f$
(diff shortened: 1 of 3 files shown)


bluescarni:
Thanks for the PR! Could you add a small test to check the new ``m_ker < 2`` throwing condition?

include/pagmo/algorithms/gaco.hpp
```diff
@@ -115,7 +115,7 @@ class PAGMO_DLL_PUBLIC gaco
      *
      * @throws std::invalid_argument if \p acc is not \f$ >=0 \f$, \p impstop is not a
      * positive integer, \p evalstop is not a positive integer, \p focus is not \f$ >=0 \f$,
-     * \p ker is not a positive integer, \p oracle is not positive, \p
+     * \p ker is not a positive integer \in [2, pop_size], \p oracle is not positive, \p
```
bluescarni, line 118: @Sceki upon construction we don't know the population size, so it looks like that the requirement ``ker <= pop_size`` should be documented for the ``evolve()`` method instead?

bluescarni (changes requested):

include/pagmo/algorithms/gaco.hpp
Sceki: @bluescarni I do check it at the evolve and throw there, but I guess you want me to modify this doc and write only >=2 here?

bluescarni: Yep it was only about the docstring. Basically, just leave ``throws if ker < 2`` in the ctor docs and add ``throws if ker is > pop_size`` in the evolve docs.

Sceki: @bluescarni Ok I pushed, have a look please :)

bluescarni: Great, thanks a lot!

PR #501, opened 2022-01-14
CMake: Fix TBB 2021 detection

jschueller:


bluescarni:
Thanks for the PR! I think eventually we just want to enforce a recent-enough version of TBB, so that we can use the CMake config-file package provided by upstream.

PR #503, opened 2022-02-28
Bump min boost version to 1.68

jschueller:
Since build with boost<1.68 fails with:

```
boost/type_traits/is_virtual_base_of.hpp:63:12: error: cannot derive from ‘final’ base ‘pagmo::detail::isl_inner<pagmo::fork_island>’ in derived type ‘boost::detail::is_virtual_base_of_impl<pagmo::detail::isl_inner_base, pagmo::detail::isl_inner<pagmo::fork_island>, boost::integral_constant<bool, true> >::boost_type_traits_internal_struct_X’
```


bluescarni:
Thanks!

Issue #505, opened 2022-03-24
Perturbation Vector in MBH is adjusted to the first problem it is called with


mlooz:
When created with default arguments, Monotonic Basin Hopping sets the perturbation vector to have size one.
When calling the evolve method with a population, the size of the perturbation vector is adjusted to fit the problem of the population. When calling the same algorithm object with a different population afterwards that has a different problem, it raises an error that the lengths of the perturbation vector does not fit the new problem. Thus, calling evolve affects the internal state of the algorithm object in an unexpected way.

An easy fix is to copy the perturbation vector in the evolve method and only adjust the size of the copy. Will do a PR for that in a couple of days.

bluescarni:
@mlooz Cheers!

bluescarni:
Fixed by #510 .

PR #510, opened 2022-04-19
Fix for perturbation in Monotonic Basin Hopping (Issue #505)

mlooz:
Creating a function-local copy of the perturbation vector in Monotonic Basin Hopping instead of marking m_perturb as volatile.


bluescarni:
@darioizzo want to take a look?

darioizzo:
Seems a valid solution ... allocating memory inside the evolve may have some consequences in nieche applications, but overall removing a mutable data member seems a step ahead.

Issue #538, opened 2023-01-25
[BUG] moead_gen is not documented


jschueller:
from the 2.19 changelog
the new moead_gen algorithm (https://github.com/esa/pagmo2/pull/508) is not mentioned in the doc
@kirbyherm

bluescarni:
#541 has been merged, closing. Please feel free to reopen @jschueller if needed.

jschueller:
looks great, thanks

PR #541, opened 2023-01-25
adding online documentation for moead_gen

kirbyherm:
addressing #538 by adding online documentation for the moead_gen algorithm.

I am not super familiar with how the html is generated but I believe these changes are sufficient.


bluescarni:
@kirbyherm everything looks in order, merging. Thanks!

PR #542, opened 2023-02-02
Fix -Wattributes warning

jschueller:
- fix a warning
- use add_test with NAME/COMMAND signautre to fix cross compilation

### DIFF
--- a/include/pagmo/archipelago.hpp
+++ b/include/pagmo/archipelago.hpp
@@ -140,7 +140,7 @@ PAGMO_DLL_PUBLIC std::ostream &operator<<(std::ostream &, migrant_handling);
 class PAGMO_DLL_PUBLIC archipelago
 {
     // Make friends with island.
-    friend class PAGMO_DLL_PUBLIC island;
+    friend class island;
 
     using container_t = std::vector<std::unique_ptr<island>>;
     using size_type_implementation = container_t::size_type;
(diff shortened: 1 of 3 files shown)


include/pagmo/archipelago.hpp
```diff
@@ -140,7 +140,7 @@ PAGMO_DLL_PUBLIC std::ostream &operator<<(std::ostream &, migrant_handling);
 class PAGMO_DLL_PUBLIC archipelago
 {
     // Make friends with island.
-    friend class PAGMO_DLL_PUBLIC island;
+    friend class island;
```
bluescarni, line 143: Could you explain what kind of warnings the ``PAGMO_DLL_PUBLIC`` attribute is giving?

As far as I remember, the original motivation for repeating ``PAGMO_DLL_PUBLIC`` here was that MSVC would complain if a class is not declared everywhere with the same exact set of attributes. But perhaps I am misremembering...

jschueller: its the Wattributes warning
I dont see any warnings in the windows compilation log

jschueller: maybe it was for an older version of msvc ?

bluescarni: Possibly... in any case, sorry for taking this long, merging this now.

PR #552, opened 2023-06-01
Write generational output

kirbyherm:
So I have a particular problem where I would like to save *ALL* the outputs from the evolutionary algorithm I'm using, not just the final population. I have hardcoded a method for implementing this, where I simply save the population at each generation to a csv file. 

Adding this PR here mainly to ask if this would be a worthwhile implementation to add across the board (either as an additional parameter, or as an additional public function similar to `set_verbosity()`.

As is, I would not pull this in but was wondering if it was worth developing further.


jschueller:
saving to a file is probably not the most flexible option, instead you can instrument the fitness method in the client code

sylvesterkaczmarek (commented):
`fopen(m_outfile.c_str(), "a")` is unchecked before the first `fprintf()`. An unwritable path, missing parent directory, or descriptor exhaustion therefore leaves `ostrm == nullptr` and makes `evolve()` invoke undefined behaviour/crash. Check the open result and throw a pagmo exception (or disable logging) before writing, with an invalid-output-path regression.

PR #553, opened 2023-06-21
De multi thread

jenyav94:
## Summary
This PR aims to introduce the batch fitness evaluation scheme to enhance the `de` algorithm.

## Overview
- This PR preserves the previous functionality as a fallback when `bfe` is not initialized.
- When bfe is initialized, all mutations are calculated prior to fitness function evaluation, which is then performed in a batch.
- The PR separates mutation and population update into distinct functions for improved code organization.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ message(STATUS "System name: ${CMAKE_SYSTEM_NAME}")
 enable_testing()
 
 # Build option: enable the test suite.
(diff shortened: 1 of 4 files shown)


bluescarni:
@jenyav94 thanks for the PR and sorry for the delay in replying.

I have asked @darioizzo to take a look as he is the author of the original algorithm, but he won't be able to spend time on this until the end of the month or so.

PR #562, opened 2023-12-12
Add initialization for potentially-used cec2014 vars

### LINKED ISSUE #561: Uninitialized variables may be used in cec2014 problem
There are two instances where variables may be used in an uninitialized state in the `src/problems/cec2014.cpp` problem:

* In `weierstrass_func`, line 498 declares `double sum2` which may be used uninitialized on line 515.
* In `oszfunc`, line 1297 declares `double xx` which may be used uninitialized on line 1317.

These prevent the cec2014 problem being built when "warnings-as-errors" compiler flags are enabled, e.g. `-Werror` for g++ or `/WX` for msvc.

This is fixed in [pmslavin/cec2014_uninitialized](/pmslavin/pagmo2/tree/cec2014_uninitialized)

pmslavin:
Fixes #561 

Adds initialization for two vars which could previously be used in an uninitialized state.

### DIFF
--- a/src/problems/cec2014.cpp
+++ b/src/problems/cec2014.cpp
@@ -496,6 +496,7 @@ void cec2014::weierstrass_func(const double *x, double *f, const unsigned nx, co
 
     unsigned i, j, k_max;
     double sum, sum2, a, b;
+    sum2 = 0.0;
     a = 0.5;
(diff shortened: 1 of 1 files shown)


bluescarni:
@pmslavin thanks!

PR #564, opened 2024-01-13
CMake: Support UNITY_BUILD

jschueller:
xref https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html


bluescarni:
Hi @jschueller

Am I understanding right that the goal here is to skip big ``.cpp`` files when doing jumbo builds?

jschueller:
hi, it's the contrary: assemble several sources into big files to speed up compilation

jschueller:
oh, indeed, this excludes one file from jumbo build because it contains a conflicting declaration with the others

bluescarni:
> oh, indeed, this excludes one file from jumbo build because it contains a conflicting declaration (cec2013.cpp & cec2014.cpp):
> 
> ```
> In file included from /home/schueller/projects/pagmo2/build/CMakeFiles/pagmo.dir/Unity/unity_4_cxx.cxx:9:
> /home/schueller/projects/pagmo2/src/problems/cec2014.cpp:51:18: error: redefinition of ‘constexpr const double pagmo::{anonymous}::E’
>    51 | constexpr double E = 2.7182818284590452353602874713526625;
>       |                  ^
> In file included from /home/schueller/projects/pagmo2/build/CMakeFiles/pagmo.dir/Unity/unity_4_cxx.cxx:5:
> /home/schueller/projects/pagmo2/src/problems/cec2013.cpp:51:18: note: ‘constexpr const double pagmo::{anonymous}::E’ previously defined here
>    51 | constexpr double E = 2.7182818284590452353602874713526625;
> ```

Oh I see, thanks for reporting the error message.

In this specific case we should really be defining constants once and for all, avoiding needless repetitions. I'll open a report about this. I am ok with merging this PR as is for the time being.

PR #566, opened 2024-02-14
cec: Use boost constants

### LINKED ISSUE #565: Avoid multiple definitions of mathematical constants
**Describe the bug**
As highlighted in #564, there are at least a few places in the source code where we are defining multiple times the same mathematical constants, leading to needless code duplication.

We should be really using always and consistently Boost's numerical constants or C++20 constants:

https://en.cppreference.com/w/cpp/numeric/constants

jschueller:
Closes #565

better use boost for consistency, also its too soon to use c++20 imho


bluescarni:
@jschueller Thanks a lot!

PR #572, opened 2024-07-04
CMake: Find Boost using CONFIG mode only

jschueller:
the module mode is deprecated in cmake>=3.30:
```
CMake Warning (dev) at cmake_modules/PagmoFindBoost.cmake:4 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

Call Stack (most recent call first):
  CMakeLists.txt:186 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```


bluescarni:
@jschueller LGTM, merging, thanks a lot!

Does this mean the official Boost CMake support is finally usable?

PR #575, opened 2024-07-26
unconstrain: Fix batch fitness

jschueller:
The original fitness has dimension nf and the unconstrained fitness nobj


bluescarni:
@jschueller am I understanding correctly here that  the issue was the initialisation of the ``y`` vector accounting only for the number of objectives and not for the number of constraints?

jschueller:
yes, exactly

bluescarni:
Ok cheers, merging.

jschueller:
@bluescarni do you know if there is a release planned this year ?

bluescarni:
> @bluescarni do you know if there is a release planned this year ?

I will try to make a new release in a few days.

bluescarni:
@jschueller new releases for pagmo/pygmo are out, sorry for taking longer than expected.

jschueller:
no problem, thanks a lot!

jschueller:
the doc does not seem up to date though; I dont see the new changelog

Issue #493, opened 2021-12-07
[BUG] gaco hangs when initial population is outside bounds


jschueller:
when some points of the initial population are outside the bounds and the objective tend to make it go outside,
gaco can hang:

```
#include <pagmo/algorithms/gaco.hpp>
#include <pagmo/population.hpp>
#include <pagmo/problems/rosenbrock.hpp>

using namespace pagmo;

struct problem1 {
    vector_double fitness(const vector_double &x) const
    {
        return {x[1]};
    }
    std::pair<vector_double, vector_double> get_bounds() const
    {
        return {{-1.5, -1.5}, {1.5, 1.5}};
    }
    vector_double::size_type get_nobj() const
    {
        return 1;
    }
};

int main()
{
  try {
    problem prob{problem1{}};

    population pop1{prob, 0, 0};
    int size = 100;
//     std::uniform_real_distribution<double> unif(-1.5, 1.5); // OK
    std::uniform_real_distribution<double> unif(-3., 3.);  // FREEZE
    std::default_random_engine re;
    for (int i=0; i<size; ++i)
    {
      vector_double x(2);
      x[0] = unif(re);
      x[1] = unif(re);
      pop1.push_back(x);
    }

    gaco user_algo1{10u};
    pop1 = user_algo1.evolve(pop1);

	}
	catch(const std::exception & exc) {
		std::cout << "ex="<<exc.what() <<std::endl;
		return 2;
	}
	return 0;
}
```
more generally, do we have to check is the population is inside the bounds before calling pagmo, or do the algorithms will manage to "jump" inside the bounds ?

bluescarni:
Any idea @Sceki ?

> more generally, do we have to check is the population is inside the bounds before calling pagmo, or do the algorithms will manage to "jump" inside the bounds ?

As I wrote in the other thread, what happens with out-of-bounds conditions is algorithm-dependent. But in any case it should never lead to a hang/crash, and if it does it's a bug.

Sceki:
I believe that the problem is that this `while` loop can be endless in case the initial population is out of bounds: https://github.com/esa/pagmo2/blob/master/src/algorithms/gaco.cpp#L855. 

I think there are two possibilities: either we avoid the while loop (i.e., the re-sampling until we satisfy the bounds) for the case in which the user decides that out-of-bounds solutions are acceptable, or we simply do not accept initial populations that are out of bounds.
[...]

PR #576, opened 2024-07-29
bug fix

Sceki:
attempted fix for https://github.com/esa/pagmo2/issues/493


Sceki:
@bluescarni this is good to go as far as I am concerned.

It should fix -> https://github.com/esa/pagmo2/issues/493. 
@jschueller it would be great if you can also check this, in case you have time

bluescarni:
Cheers!

PR #577, opened 2024-07-29
problem: Mention integer/continuous variables order

### LINKED ISSUE #496: [BUG] ambiguous problem integer/continous components layout documentation
for integer problems, the documentation is unclear about how integer and continuous components must be distinguished:
https://esa.github.io/pagmo2/docs/cpp/problem.html#_CPPv4NK5pagmo7problem7get_nixEv

I figured out from the pygmo doc that integers must be tucked at the end after the continuous components:
https://esa.github.io/pygmo2/tutorials/coding_udp_minlp.html

I suggest to mention that in the get_nix method doc

jschueller:
Closes #496


bluescarni:
Thanks @jschueller !

PR #580, opened 2024-08-20
Update compass_search.cpp

Tas-I-te:
0 <= stop range < start range


Tas-I-te:
The description of document should be also modified

PR #583, opened 2024-09-25
CMake: Find dependencies in QUIET mode from pagmo-config

jschueller:
only when pagmo is used as a client library, not when configuring pagmo itself


bluescarni:
@jschueller thanks for the PR, LGTM.

Regarding the ``clang>=17`` issues on OSX, this seems hairy and I will try to explain to the best of my understanding what is going on.

The issue is a ``std::bad_cast`` exception being thrown during (de)serialisation of problems/algorithms/etc. In pagmo, we employ the ``boost::serialization`` library, which we use to (de)serialize derived classes via pointers to their bases. For this to work, RTTI is used, and ``boost::serialization`` needs to register a caster between base and derived classes in some internal global state, and to perform ``std::dynamic_cast``s at runtime.

This has been working great for years in pagmo, but somehow, for reasons I cannot understand, it has stopped working since the clang 17 upgrade. I have run in the same issue in another project of mine, where I also have started pinning clang on osx:

https://github.com/conda-forge/heyoka.py-feedstock/blob/main/recipe/conda_build_config.yaml

According to my testing in this other project, the issue arises only on ``x86_64``, apparently ``osx_arm64`` is fine.

It looks like this is a quite difficult issue to debug and pin down. Perhaps someone with deep toolchain knowledge (pinging @isuruf and @h-vetinari ) could chime in...

bluescarni:
Ok maybe not so difficult to diagnose after all. See:

https://releases.llvm.org/17.0.1/tools/clang/docs/ReleaseNotes.html#c-language-changes

And:

https://github.com/llvm/llvm-project/issues/71196

I will try the suggested workaround...

bluescarni:
@jschueller would you like to rebase/merge ``master`` back into this PR so I can merge this?

jschueller:
done

bluescarni:
Cheers!

Issue #579, opened 2024-08-10
[BUG] Many tests fail with unknown location(0): fatal error:


yurivict:
**Describe the bug**

Most or all failures look like this:
```
$ tests/unconnected 
Running 2 test cases...
unknown location(0): fatal error: in "basic_test": boost::archive::archive_exception: unregistered void cast N5pagmo6detail10topo_innerINS_11unconnectedEEE<-N5pagmo6detail15topo_inner_baseE
/usr/ports/science/pagmo2/work/pagmo2-2.19.1/tests/unconnected.cpp(55): last checkpoint

*** 1 failure is detected in the test module "unconnected"

$ tests/translate 
Running 8 test cases...
unknown location(0): fatal error: in "translate_serialization_test": boost::archive::archive_exception: unregistered void cast N5pagmo6detail10prob_innerINS_9translateEEE<-N5pagmo6detail15prob_inner_baseE
/usr/ports/science/pagmo2/work/pagmo2-2.19.1/tests/translate.cpp(105): last checkpoint: "translate_serialization_test" test entry

*** 1 failure is detected in the test module "translate_test"

$ tests/unconstrain 
Running 8 test cases...
unknown location(0): fatal error: in "unconstrain_serialization_test": boost::archive::archive_exception: unregistered void cast N5pagmo6detail10prob_innerINS_11unconstrainEEE<-N5pagmo6detail15prob_inner_baseE
/usr/ports/science/pagmo2/work/pagmo2-2.19.1/tests/unconstrain.cpp(178): last checkpoint: "unconstrain_serialization_test" test entry

*** 1 failure is detected in the test module "unconstrain_test"
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: FreeBSD 14.1
 - Installation method: port
 - Version: 2.19.1

jschueller:
seems to also happen on osx, which also uses clang, which clang version are you using ?
and boost version ?

PR #585, opened 2024-09-26
CMake: Workaround clang17+ / lto bug

jschueller:


bluescarni:
@jschueller thanks for the fast PR!

As far as I understand it, for the LLVM bug to trigger, link-time optimisation must be enabled - at least judging from this bugreport:

https://github.com/llvm/llvm-project/issues/71196

I was thus wondering if it would make sense to just disable LTO if clang>=17 is detected, rather than injecting the ``-fno-assume-unique-vtables``. I am not sure though, what do you think?

jschueller:
lets keep the new flag, but only add it if IPO is enabled

jschueller:
will close #579

bluescarni:
@jschueller sorry for getting back to this only now...

As far as I understand from #586, switching off IPO actually did not prevent the bug from manifesting right? If that is the case and it is ok with you, I would slightly edit this PR to enable ``-fno-assume-unique-vtables`` regardless of whether or not IPO is enabled.

bluescarni:
@jschueller thank you!

PR #589, opened 2024-11-01
Add missing Ipopt return codes.

saverkin:
Maximum_CpuTime_Exceeded return code was introduced in at least Ipopt 3.10.4 (2013-05-05). Maximum_WallTime_Exceeded return code was added in Ipopt 3.14.0 (2021-06-15). Please, see Ipopt changelog for details: https://coin-or.github.io/Ipopt/md_ChangeLog.html. Add missing Ipopt return codes.

Without these return codes, the pagmo2 throws an exception when any of these conditions is satisfied.


darioizzo:
Seems good to me. The change should not break use with previous ipopt versions as the specific new return code will simply never trigger.. No need to add a test on this. 

CI not passing is more concerning, but it run ages ago. @saverkin can you update your branch and trigger new builds?

saverkin:
Thanks for the quick response! I updated my branch. As for the failed CI, as far as I remember it complained about missing Ubuntu image or something similar.

saverkin:
Hello @bluescarni, it seems like all tests have passed.

PR #590, opened 2025-01-13
CMake: Bump require CMake version

jschueller:
Avoids the cmake<3.10 deprecation message and simplifies some code
cmake 3.13 is still old enough (debian 10)


jschueller:
I'm not sure I can help for this one sorry

PR #600, opened 2025-04-10
Drop ciso646 deprecated header

jschueller:
Else I get a warning with gcc 15:
```
/usr/include/c++/15/ciso646:46:4: warning: #warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros" [-Wcpp]
   46 | #  warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros"
```

I checked including vector instead also yields the _LIBCPP_VERSION macro with clang/libc++

Also fixes the doc CI job

### DIFF
--- a/config.hpp.in
+++ b/config.hpp.in
@@ -31,7 +31,7 @@ see https://www.gnu.org/licenses/. */
 
 // NOTE: include this so that we can
 // detect _LIBCPP_VERSION below.
-#include <ciso646>
+#include <vector>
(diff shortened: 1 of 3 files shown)


bluescarni:
Thanks @jschueller !

The idea with ciso646 was to include the lightest possible header that would allow us to detect the standard library version. Had no idea this was being deprecated...

PR #610, opened 2025-11-14
CMake: Honor PackageName_FIND_QUIETLY

jschueller:


jschueller:
@darioizzo what do you think of this one ?

darioizzo:
This is a good, narrowly scoped PR, thanks @jschueller. Merging.

PR #613, opened 2026-01-06
island: Fix get_ptr

### LINKED ISSUE #584: [BUG] island.hpp:226:16: error: cannot initialize return object of type 'void *'
**Describe the bug**

compiling the island test with clang >= 17 on linux fails with:
```
In file included from /tmp/pagmo2/tests/island.cpp:56:
/tmp/pagmo2/include/pagmo/island.hpp:226:16: error: cannot initialize return object of type 'void *' with an rvalue of type 'const pagmo::thread_island *'
  226 |         return &m_value;
      |                ^~~~~~~~
/tmp/pagmo2/include/pagmo/island.hpp:1057:20: note: in instantiation of member function 'pagmo::detail::isl_inner<const pagmo::thread_island>::get_ptr' requested here
 1057 |         auto isl = dynamic_cast<detail::isl_inner<T> *>(m_ptr->isl_ptr.get());
      |                    ^
/tmp/pagmo2/tests/island.cpp:536:21: note: in instantiation of function template specialization 'pagmo::island::extract<const pagmo::thread_island>' requested here
  536 |     BOOST_CHECK(isl.extract<const thread_island>() == nullptr);
      |
```

**To Reproduce**

just build master on latest ubuntu:
```
FROM ubuntu:noble
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get -y update && apt-get -y install cmake clang libtbb-dev libeigen3-dev git libboost-serialization-dev libboost-test-dev g++
WORKDIR /tmp
RUN git clone --depth 1 https://github.com/esa/pagmo2.git
RUN cmake -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DCMAKE_UNITY_BUILD=ON -DPAGMO_WITH_EIGEN3=ON -DPAGMO_BUILD_TESTS=ON -S pagmo2 -B build 
RUN cmake --build build --parallel 8
RUN ctest --test-dir build
```

jschueller:
Closes #584

### DIFF
--- a/doc/sphinx/install.rst
+++ b/doc/sphinx/install.rst
@@ -22,7 +22,7 @@ The pagmo C++ library has the following **mandatory** dependencies:
 
 Additionally, pagmo has the following **optional** dependencies:
 
-* `Eigen3 <http://eigen.tuxfamily.org/index.php?title=Main_Page>`__ (which is required
+* `Eigen3 <https://libeigen.gitlab.io/>`__ (which is required
   by some algorithms, e.g., :cpp:class:`pagmo::cmaes`), version 3.3 or later,
 * `NLopt <https://nlopt.readthedocs.io/en/latest/>`__ (which is required by
   the :cpp:class:`pagmo::nlopt` wrapper), version 2.6 or later,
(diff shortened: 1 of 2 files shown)


jschueller:
hi @bluescarni , any thoughts on this ?

darioizzo:
This is a good catch and in general we may improve idiomatically this in the future, but current GCC/Clang and runners all pass, that usually means the problematic instantiation path is no longer being hit in practice. @jschueller fix makes sense and is understandable, but 

It is semantically noisier and suggests const-correctness is being bypassed for a reason, when it is not.
It can hide real constness bugs if this pattern spreads.
It reduces readability and maintainability without changing behavior for this case.

WIll not merge for now. But lets keep this in mind for the future.

jschueller:
can you cherry-pick the doc fix though ?

darioizzo:
I think its already fixed with the last main CI refresh?

Issue #582, opened 2024-09-17
[FEATURE] Approximating Hypervolume / Hypervolume Contributions by polar coordinates


CoolRunning:
**Is your feature request related to a problem? Please describe.**
The monte-carlo based hypervolume approximation in pagmo `bf_fpras` / `bf_approx` is kinda outdated by now. There exists alternatives that deploy some number-theoretic tricks and coordinate transformations that makes them particular robust for certain geometries (e.g. highly convex or concave point sets) in the sense that they select the "correct" least contributor with higher chance. Finding the least contributor is critical for many-objective optimization if the number of objectives goes high (let's say >8). 

**Describe the solution you'd like**
It would be cool to see the algorithm from Deng and Zhang (see context) implemented in pagmo. The idea of this algorithm is to approximate the hypervolume by integrating the lengths of rays coming from from the reference point towards the front, using some sampling of polar coordinates/angles. The implementation might be moderately complex, but probably doable with some commitment. 

**Describe alternatives you've considered**
To be fair, the approximation right now in pygmo is not bad and does its job. This is more a suggestion for a hypervolume related project, that can be put forward for interested individuals or groups. For this to be useful, many-objective optimization problems (>8 objectives) would need to be of practical concern. Otherwise, this is mostly of academic interest.

**Additional context**
Deng, J. and Zhang, Q. Approximating hypervolume and hypervolume contributions using polar coordinate. IEEE Transactions on Evolutionary Computation, 2019. [[link](https://ieeexplore.ieee.org/document/8625504)]

PR #615, opened 2026-01-28
Add: Initial implementation of polar_approx and unit tests

mariam851:
Implemented a new hypervolume approximation algorithm using Polar Coordinates (Deng & Zhang approach) as requested in issue #582. This method improves sampling efficiency in high-dimensional spaces.

2. Technical Changes
Core Implementation: Added polar_approx class in src/utils/hv_algos/.

Ray Generation: Implemented generate_random_ray() using std::normal_distribution to ensure uniform sampling on the unit hypersphere surface.

Normalization: Added logic to ensure all generated rays are Unit Vectors and stay within the positive orthant.

3. Build & Test
CMake: Updated tests/CMakeLists.txt to register the new test case.

Unit Tests: Added a new test file to verify the mathematical properties of the generated rays (norm ≈ 1.0 and values ≥ 0).


sylvesterkaczmarek (commented):
Both the new implementation and test include `pagmo/utils/hv_algos/hv_polar_approx.hpp`, but that header is neither in `master` nor added by this PR, so the build fails before the new code can compile. Add the public header declaring `polar_approx` and keep a build/install test covering that include path.

darioizzo:
closing for lack of action. happy to reconsider.

PR #616, opened 2026-02-26
Possible 50% performance improvement for NSGA-II on MSVC 

jonas-s-s-s:
Hello everyone,
as part of my Master's Thesis (Faculty of Applied Sciences, University of West Bohemia), I have been investigating the poor performance of pagmo2's implementation of NSGA2 when compiled with MSVC. The changes that I've made decrease the execution time by about 50% (MSVC 19.44.35217). This improvement was possible due to the inefficiency of `std::isnan` in the MSVC Runtime Library.
Other compilers such as (Clang and GCC) benefit mainly from the memory optimizations, decreasing the execution time by about 15%. 

The main changes in this PR include:
- Reimplemented` less_than_f`, `greater_than_f`, `equal_to_f` using C++20 starship operator
	- 30% performance improvement on MSVC
- Inlined `pareto_dominance()`
	- Eliminated high function call overhead in hot path loops
- Added `fast_non_dominated_sorting_buffered()` and `select_best_N_mo_buffered()`
	- Prevents repeated destruction an re-allocation of std::vectors that are used as a return value
	- Obviously this "duplicates" the logic of these functions, however making a new function which has the pre-allocated struct passed to it as parameter was the only option since these functions aren't part of any class we can't use class members for this 

Detailed explanation together with the code that was used for profiling can be found here: https://github.com/jonas-s-s-s/Pagmo-NSGA2-Improve-Test

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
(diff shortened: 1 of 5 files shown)


sylvesterkaczmarek (commented):
`pareto_dominance()` is now declared `inline` in the public header, but its definition remains only in `multi_objective.cpp`. Downstream translation units therefore see an inline function with no definition and can fail to link; keep the exported non-inline API, or move the complete inline definition into the header and add an external-consumer link test.

jonas-s-s-s:
> `pareto_dominance()` is now declared `inline` in the public header, but its definition remains only in `multi_objective.cpp`. Downstream translation units therefore see an inline function with no definition and can fail to link; keep the exported non-inline API, or move the complete inline definition into the header and add an external-consumer link test.
> 

Yeah I should probably fix this before I forget what this was about. iirc the most important thing is not using std::isnan on MSVC as it's the function that's slowing it down the most. The new functions that accept vector references as a parameter are more like a hack to gain extra 5% performance, it was the easiest way to prevent the wasteful re-allocation of this memory.

PR #617, opened 2026-03-26
refactor: remove unused variable from `nsga2::evolve`

vil02:


jschueller:
should be good to go @darioizzo

PR #623, opened 2026-05-21
fix: resolve clang >= 17 compilation error in extract<const T>()

### LINKED ISSUE #584: [BUG] island.hpp:226:16: error: cannot initialize return object of type 'void *'

jschueller:
The non-const get_ptr() in UDx inner classes tries to return &m_value as void*, but when T is const-qualified this drops const without an explicit cast, which clang >= 17 rejects. Use std::remove_const_t<T> for the m_value member so &m_value is always a non-const pointer. This avoids const_cast entirely and applies uniformly across all UDx types (island, problem, algorithm, topology, bfe, s_policy, r_policy).

Closes #584.

### DIFF
--- a/include/pagmo/algorithm.hpp
+++ b/include/pagmo/algorithm.hpp
@@ -366,7 +366,7 @@ private:
     }
 
 public:
-    T m_value;
+    std::remove_const_t<T> m_value;
 };
 
 } // namespace detail
@@ -529,8 +529,8 @@ public:
 #if defined(PAGMO_PREFER_TYPEID_NAME_EXTRACT)
         return detail::typeid_name_extract<T>(*this);
 #else
-        auto p = dynamic_cast<const detail::algo_inner<T> *>(ptr());
-        return p == nullptr ? nullptr : &(p->m_value);
+        auto p = dynamic_cast<const detail::algo_inner<uncvref_t<T>> *>(ptr());
+        return p == nullptr ? nullptr : reinterpret_cast<const T *>(&(p->m_value));
 #endif
     }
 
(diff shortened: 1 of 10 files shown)


sylvesterkaczmarek (commented):
`m_value` is now non-const, but `extract<const T>()` still casts to `algo_inner<const T>`. Construction stores `algo_inner<uncvref_t<T>>`, so a const-qualified extraction still returns `nullptr` instead of matching the underlying UDA as the API docs promise. Please strip cv in the cast target too and add an `extract<const T>()` regression test.

jschueller:
done, thanks

sylvesterkaczmarek (approved):
Rechecked current 19f11d36. The const-qualified extraction issue I raised is resolved: the stored UDA is unqualified, extract<T>() now casts against uncvref_t<T>, and the typeid-name path applies the same normalization. The regressions verify extract<const T>() and is<const T>() return the same underlying object rather than nullptr. Current CI is green. No remaining blocker from my review.

PR #624, opened 2026-05-21
mahco: filter initial population to verify constraints

### LINKED ISSUE #525: [BUG] mhaco asserts if initial population does not verify constraints
if the initial population does not verify constraints (through unconstrain) mhaco asserts:
```
IndexError: 
function: assert_minimisation
where: /tmp/pagmo2-2.18.0/src/utils/hv_algos/hv_algorithm.cpp, 256
what: Reference point is invalid: another point seems to be outside the reference point boundary, or be equal to it:
```
kind of similar to #493

jschueller:
Filter out non-finite fitness values (e.g., from death-penalty unconstrain) before hypervolume computation to prevent assert_minimisation failure when the reference point equals infeasible points (both == DBL_MAX).

Closes #525

### DIFF
--- a/src/algorithms/maco.cpp
+++ b/src/algorithms/maco.cpp
@@ -30,6 +30,7 @@ see https://www.gnu.org/licenses/. */
 #include <cmath>
 #include <iomanip>
 #include <iostream>
+#include <limits>
 #include <iterator>
 #include <numeric>
 #include <random>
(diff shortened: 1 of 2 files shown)


Sceki:
Hi @jschueller , I just noticed this: I am wondering, shouldn't this be done at the level of the hypervolume algorithm, instead of specifically for MHACO? I mean if tomorrow another algorithm will use the hypervolume codebase it might display the same issue, right?

sylvesterkaczmarek (commented):
When first-front extremities are forced into the tail of `sol_archive`, the paired fitness row is populated from `sol_archive[i_f]` instead of the row just overwritten at `m_ker - 1 - i_f`. `sol_archive_fit` can therefore disagree with the archive before `ideal(sol_archive_fit)` is computed. Copy `border_fits[i_f]` directly (or read the matching destination row) and add an extremity-replacement regression.

jschueller:
@sylvesterkaczmarek done, thanks

@Sceki Valid concern, but MHACO-level is correct: hypervolume contract is finite, strictly-dominated points — silently filtering DBL_MAX/inf inside it would hide caller bugs. Better keep hypervolume strict (throw with clear message on non-finite) and let algorithms that use death-penalty (only unconstrain) filter explicitly as now done — documents intent and fails fast for future users.

Sceki:
@jschueller perhaps I am misunderstanding, but is this not triggered because the hypervolume returns a non-finite number? 

Generally speaking I think it would be useful to have an example of what is exactly the issue and what is triggering it

jschueller:
No, hypervolume does not return non-finite – hv.refpoint(0.1) saturates to DBL_MAX when the front contains death-penalty DBL_MAX points, then hv.contributions(ref) throws assert_minimisation: Reference point is invalid (point equal/outside). Filtering at MACO level keeps hypervolume strict; reproducer is maco_constrained_test.

There is a new test to verify the new behavior

sylvesterkaczmarek (approved):
The extremity replacement now writes `border_fits[i_f]` into the matching `sol_archive_fit[m_ker - 1 - i_f]` row, so the decision and fitness archives remain paired. The new extremity regression checks both endpoints and the ideal across iterations. This resolves the mismatch I raised.

darioizzo:
@Sceki reading the above I tend to agree with @jschueller ... can you assess if there are any othe hidden consequences in the algo evolve, else I would say its a safe change.

Sceki:
Hi @jschueller , all right, I see now, thanks for clarifying.

So having a better look at it, I think the `evolve` behavior of `mhaco` is only affected by one of your changes:

* the new filter (`if (!std::isfinite(val) || val == std::numeric_limits<double>::max()) [...]`), which skips the infeasible points before computing the hypervolume

correct me if I am wrong, but with the proposed fix, if at the very first generation you don't have enough feasible individuals, then the archive will be filled with initialized values, compromising the behavior of the algo?

Perhaps it would be best to still keep these individuals at the end of the archive to prevent this from happening?

jschueller:
done, padding with skipped individuals

Sceki:
Thanks. @jschueller maybe it would also be worth adding a test with all infeasible individuals to check this new behavior?

Sceki:
@jschueller could you please also trigger in the tests -> lines 392-393, and lines 462-467?

jschueller:
ok, done

Sceki (approved):

darioizzo:
Thanks everyone, merging. @jschueller please for next ones avoid force commits as it hides visibility on the conversations and reviews.

PR #625, opened 2026-05-21
cmaes/xnes: enforce problem bounds by default

### LINKED ISSUE #494: [BUG] cmaes & xnes do not honor bounds
it seems cmaes (and xnes) does not care about our bounds as the final population ends up way outside eventhough the initial population is inside the bounds:

```
#include <pagmo/algorithms/cmaes.hpp>
#include <pagmo/population.hpp>
#include <pagmo/problems/rosenbrock.hpp>

using namespace pagmo;

struct problem1 {
    vector_double fitness(const vector_double &x) const
    {
        double x1=x[0];
        double x2=x[1];
        return {x2};
//         return {1.+100.*(x2-x1*x1)*(x2-x1*x1)+(1.-x1)*(1.-x1)};
    }
    std::pair<vector_double, vector_double> get_bounds() const
    {
        return {{-1.5, -1.5}, {1.5, 1.5}};
    }
    vector_double::size_type get_nobj() const
    {
        return 1;
    }
};

int main()
{
  try {
    problem prob{problem1{}};

    population pop1{prob, 0, 0};
    int size = 100;
    std::uniform_real_distribution<double> unif(-1.5, 1.5);
    std::default_random_engine re;
    for (int i=0; i<size; ++i)
    {
      vector_double x(2);
      x[0] = unif(re);
      x[1] = unif(re);
      pop1.push_back(x);
    }

    cmaes user_algo1{10u};
    pop1 = user_algo1.evolve(pop1);
    for (int i=0; i<size; ++i)
    {
      vector_double x = pop1.get_x()[i];
      std::cout << "x1="<<x[0]<<" x2="<<x[1] <<std::endl;
    }

	}
	catch(const std::exception & exc) {
		std::cout << "ex="<<exc.what() <<std::endl;
		return 2;
	}
	return 0;
}


```

jschueller:
Change the default value of the `force_bounds` constructor parameter from `false` to `true` so that both algorithms clamp new individuals to the box bounds during evolution, matching the behavior of all other pagmo algorithms.

Closes #494

### DIFF
--- a/include/pagmo/algorithms/cmaes.hpp
+++ b/include/pagmo/algorithms/cmaes.hpp
@@ -118,13 +118,13 @@ public:
      * @param xtol stopping criteria on the f tolerance (default is 1e-6)
      * @param memory when true the adapted parameters are not reset between successive calls to the evolve method
      * @param force_bounds when true the box bounds are enforced. The fitness will never be called outside the bounds
-     but the covariance matrix adaptation  mechanism will worsen
+     but the covariance matrix adaptation  mechanism will worsen (default is true)
      * @param seed seed used by the internal random number generator (default is random)
      * @throws std::invalid_argument if cc, cs, c1 and cmu are not in [0, 1]
      */
-    cmaes(unsigned gen = 1, double cc = -1, double cs = -1, double c1 = -1, double cmu = -1, double sigma0 = 0.5,
-          double ftol = 1e-6, double xtol = 1e-6, bool memory = false, bool force_bounds = false,
-          unsigned seed = pagmo::random_device::next());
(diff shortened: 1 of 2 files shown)


darioizzo:
I am not sure I want to enforce bounds as a pagmo udas requirement. In the specific for CMA-ES bounds enforcement breaks the algorithmic mathematical coherence and can result in degraded performance as well as in degraded algorithmic performances. As is the default values are at least reproducing the known peformances of CMA-ES as published in the ref paper. @jschueller 

It is probably a superior thing to aim for that default values are reproducing the algorithmic behaviour reported or expeced from the literature.

jschueller:
ok, no problem

PR #626, opened 2026-05-21
archipelago: auto-sync topology vertex count in set_topology to preve…

### LINKED ISSUE #606: [BUG] set_topology causing crash.
**Describe the bug**
If set the 
```
archipelago archi{ 16u, algo, prob, 20u };
archi.set_topology(pagmo::topology(pagmo::ring()));
```
then program crash. remove the line ok.

**To Reproduce**
Code:
```
#include <iostream>

#include <pagmo/algorithm.hpp>
#include <pagmo/algorithms/sade.hpp>
#include <pagmo/algorithms/pso.hpp>
#include <pagmo/archipelago.hpp>
#include <pagmo/problem.hpp>
#include <pagmo/problems/schwefel.hpp>
#include <pagmo/topologies/ring.hpp>
using namespace pagmo;

struct CustomSphere {
	CustomSphere(unsigned int dim = 30) : m_dim(dim) {}

	std::vector<double> fitness(const std::vector<double>& x) const {
		double sum = 0.0;
		for (const auto& xi : x) {
			sum += xi * xi;
		}
		return { sum };
	}

	std::pair<std::vector<double>, std::vector<double>> get_bounds() const {
		std::vector<double> lb(m_dim, -5.0);
		std::vector<double> ub(m_dim, 5.0);
		return { lb, ub };
	}

	unsigned int get_nobj() const {
		return 1;
	}

	std::string get_name() const {
		return "Custom Sphere Function";
	}

private:
	unsigned int m_dim;
};


int main()
{
	// 1 - Instantiate a pagmo problem constructing it from a UDP
	// (i.e., a user-defined problem, in this case the 30-dimensional
	// generalised Schwefel test function).
	problem prob{ CustomSphere(30) };

	// 2 - Instantiate a pagmo algorithm (self-adaptive differential
	// evolution, 100 generations).
	algorithm algo{ pagmo::pso(100) };

	algo.set_verbosity(5);
	// 3 - Instantiate an archipelago with 16 islands having each 20 individuals.
	archipelago archi{ 16u, algo, prob, 20u };
	archi.set_topology(pagmo::topology(pagmo::ring()));

	// 4 - Run the evolution in parallel on the 16 separate islands 10 times.
	archi.evolve(10);

	// 5 - Wait for the evolutions to finish.
	archi.wait_check();

	// 6 - Print the fitness of the best solution in each island.
	for (const auto& isl : archi) {
		std::cout << isl.get_population().champion_f()[0] << '\n';
	}
}
```

**Expected behavior**

**Screenshots**

 - OS: [e.g. Linux] windows
 - Installation method: [e.g. conda, source] vcpkg
 - Version: [e.g. 2.15] 2.19.1

**Additional context**

jschueller:
…nt crash

Calling set_topology() with a freshly constructed topology (e.g. pagmo::topology(pagmo::ring())) would replace the archipelago's topology with one having 0 vertices, while the archipelago may already contain islands. Subsequent evolve() would then crash when get_connections() is called with an out-of-bounds vertex index.

Fix by adding a num_vertices() method to the topology class and modifying set_topology() to automatically call push_back() until the topology has enough vertices for all existing islands.

Fixes #606

### DIFF
--- a/include/pagmo/topology.hpp
+++ b/include/pagmo/topology.hpp
@@ -118,6 +118,21 @@ using bgl_graph_t
 
 #endif
 
+// Detect the num_vertices() method.
+template <typename T>
+class has_num_vertices
+{
+    template <typename U>
+    using num_vertices_t = decltype(std::declval<const U &>().num_vertices());
+    static const bool implementation_defined = std::is_same<std::size_t, detected_t<num_vertices_t, T>>::value;
+
+public:
+    static const bool value = implementation_defined;
+};
+
+template <typename T>
+const bool has_num_vertices<T>::value;
+
 // Detect the to_bgl() method.
 template <typename T>
 class has_to_bgl
@@ -171,6 +186,7 @@ struct PAGMO_DLL_PUBLIC_INLINE_CLASS topo_inner_base {
     virtual std::string get_name() const = 0;
     virtual std::string get_extra_info() const = 0;
     virtual std::pair<std::vector<std::size_t>, vector_double> get_connections(std::size_t) const = 0;
+    virtual std::size_t num_vertices() const = 0;
     virtual void push_back() = 0;
     virtual bgl_graph_t to_bgl() const = 0;
     virtual std::type_index get_type_index() const = 0;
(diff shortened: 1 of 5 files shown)


sylvesterkaczmarek (commented):
The `num_vertices_impl()` fallback makes `set_topology()` non-terminating for existing custom topology UDAs that implement `push_back()` but not the new optional `num_vertices()`: every loop iteration still observes zero vertices. Either make vertex count part of the required topology concept or make the fallback track successful pushes, and add a legacy custom-topology regression.

jschueller:
@sylvesterkaczmarek, done, thanks

sylvesterkaczmarek (approved):
Rechecked current 4ce7f086. The legacy-UDT nontermination issue I raised is resolved: topologies without num_vertices() now track successful push_back() calls in the type-erased wrapper, and that fallback count is preserved across clone/serialization. set_topology() computes the missing count once and the regression exercises a legacy custom topology through evolve. Current CI is green. No remaining blocker from my review.

darioizzo:
Thanks for the fix @jschueller looks good to merge.

The patch fixes the crash caused by replacing an archipelago topology with a freshly constructed topology that has no vertices. It synchronizes the topology with the existing islands, preserves compatibility with built-in and legacy UDTs, and adds regression coverage for both cases. The focused archipelago and topology tests pass.

One minor residual risk is that a legacy UDT without num_vertices() could be initialized with pre-existing vertices. The fallback counter cannot detect those vertices and could over-add during synchronization. This does not affect the provided topologies or UDTs implementing num_vertices(), and it is outside the tested/common path.

merging

sylvesterkaczmarek:
Thanks for the review and merge. The legacy-topology fallback was the main compatibility edge I was concerned about, so I appreciate the regression coverage and final check.

PR #627, opened 2026-05-21
hypervolume: fix flaky floating point comparisons tests

### LINKED ISSUE #516: Comparing floating point numbers in tests : comparing difference instead of using ==
Hello,

In the hypervolume tests, some double are compared using ==, which can lead to test failures as it is safer to compare them by checking that their difference is small. 
In the attached patch I changed some of these tests so that the failing tests now pass, but there remains some other such tests to change I think.

Best,
Pierre

[check_equality_floating.txt](https://github.com/esa/pagmo2/files/8879299/check_equality_floating.txt)

jschueller:
Replace equality checks (==) with BOOST_CHECK_SMALL using a 1e-12 tolerance for computed double values throughout the hypervolume test suite, fixing test failures on platforms with differing floating point precision.

Closes #516

### DIFF
--- a/tests/hypervolume.cpp
+++ b/tests/hypervolume.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2017-2021 PaGMO development team
+/* Copyright 2017-2026 PaGMO development team
 
 This file is part of the PaGMO library.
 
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
`assertContribs()` still compares `hv.contributions(ref)` to `answers` with exact vector equality in all three configurations. Those values are computed doubles, so the same platform-dependent rounding this PR is fixing can still fail here. Compare the contribution vectors elementwise with the same tolerance and add a regression that exercises this helper.

jschueller:
@sylvesterkaczmarek done

sylvesterkaczmarek (approved):
`assertContribs()` now compares every contribution/exclusive value elementwise with tolerance in all copy/no-copy configurations instead of relying on exact vector equality. That addresses the platform-rounding failure I raised.

darioizzo:
merging this, affecting tests only and relaxng some possible platform float maddness.

PR #628, opened 2026-05-21
PSO: fix memory mode to preserve full internal state across evolve calls

### LINKED ISSUE #486: [BUG] Consecutive calls to PSO is not necessarily the same of running with multiple generations

jschueller:
When memory=true, consecutive evolve() calls with gen=1 should produce the same trajectory as a single call with gen=N. The previous partial fix only preserved particle velocities (m_V), but not the rest of the internal state (current positions X, local bests lbX/lbfit, topology, etc.), causing divergent behaviour between repeated gen=1 calls and a single gen=N call.

This commit stores and restores the full internal PSO state via a memory struct (V, X, lbX, fit, lbfit, best_fit, neighb, best_neighb) wrapped in boost::optional, replacing the incomplete mutable m_V approach.

Fix contributed by tarcisiofischer (PR #487), applied to both pso and pso_gen with serialization support.

Closes #486

### DIFF
--- a/include/pagmo/algorithms/pso.hpp
+++ b/include/pagmo/algorithms/pso.hpp
@@ -33,6 +33,8 @@ see https://www.gnu.org/licenses/. */
 #include <tuple>
 #include <vector>
 
+#include <boost/optional.hpp>
+
 #include <pagmo/algorithm.hpp>
 #include <pagmo/detail/visibility.hpp>
 #include <pagmo/population.hpp>
@@ -223,6 +225,20 @@ public:
     }
 
 private:
+    struct memory {
+        std::vector<vector_double> m_V;
+        std::vector<vector_double> m_X;
+        std::vector<vector_double> m_lbX;
+        std::vector<vector_double> m_fit;
+        std::vector<vector_double> m_lbfit;
+        vector_double m_best_fit;
+        std::vector<std::vector<population::size_type>> m_neighb;
+        vector_double m_best_neighb;
+
+        template <typename Archive>
+        void serialize(Archive &, unsigned);
+    };
+
     // Object serialization
     friend class boost::serialization::access;
     template <typename Archive>
@@ -254,8 +270,7 @@ private:
     unsigned m_neighb_param;
     // memory
     bool m_memory;
-    // particles' velocities
-    mutable std::vector<vector_double> m_V;
+    mutable boost::optional<pso::memory> m_memory_data;
 
     mutable detail::random_engine_type m_e;
     unsigned m_seed;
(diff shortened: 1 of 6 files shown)


darioizzo:
Merging this, thanks for the contribution. With the added test we should be safe on intended behaviour.

PR #629, opened 2026-05-21
polynomial mutation: remove redundant delta2 computation

### LINKED ISSUE #470: [Feature Change] Possible changes in Polynomial Mutation
In the polynomial mutation code, we have (Notice the comment in the code)

```c++
void polynomial_mutation_impl(vector& child .....)
{

for(....){
            y = child[j];
            yl = lb[j];
            yu = ub[j];
            delta1 = (y - yl) / (yu - yl); //Normalised distance from lower bound.
            delta2 = (yu - y) / (yu - yl); //Normalised distance from upper bound.
         if( rnd < 0.5)
                xy = 1. - delta1;   //Possibly naive
               ....................
}
}
```

The problem I find is when we do ```xy = 1. - delta1```, is unnecesary because ```1. - delta1``` is **equivalent** to ```delta2```. 
Simple proof:
```
delta1 + delta2 = ( (y - yl) + (yu - y) ) / (yu - yl) 
=> (yu - yl)/(yu - yl)
=>delta1 + delta 2 = 1
=> delta2 = 1 - delta1
```

jschueller:
delta1 + delta2 = 1 trivially, so delta2 and 1 - delta1 are equivalent. Drop delta2 entirely and compute xy from delta1 directly in both branches.

Closes #470

### DIFF
--- a/src/utils/genetic_operators.cpp
+++ b/src/utils/genetic_operators.cpp
@@ -156,7 +156,7 @@ void polynomial_mutation_impl(vector_double &child, const std::pair<vector_doubl
     const auto &lb = bounds.first;
     const auto &ub = bounds.second;
     // declarations
-    double rnd, delta1, delta2, mut_pow, deltaq;
+    double rnd, delta1, mut_pow, deltaq;
     double y, yl, yu, val, xy;
     // Random distributions
     std::uniform_real_distribution<> drng(0., 1.); // to generate a number in [0, 1)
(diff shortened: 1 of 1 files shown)


jschueller:
@darioizzo this should be pretty trivial to review too

darioizzo:
@jschueller Mathematically they are indeed equivalent, they also are equally efficient as this really will not be a bottleneck, but I prefer code readability over all (the reason it was written like this is likely to follow the originating paper / pseudocode), and am thus inclined to leave as is.

jschueller:
fair enough

PR #630, opened 2026-05-21
fix sbx crossover and nsga2 initial population bounds checks

### LINKED ISSUE #567: No "NaN/infinity" verification in sbx crossover implementation AND no bounds verification for NSGA-II's initial population
**Describe the bug**
There is no NaN/infinity verification in the SBX Crossover implementation. This makes it possible for the crossover to assign NaN/infinity values to genes. 

An example of how this seems to happen is when the population is set with genes that don't respect their bounds in NSGA-II (and maybe other genetic algorithm implementation). There should be a verification beforehand that all those genes respect the bounds set.

**Expected behavior**
**For NSGA-II**
1. An error message saying that the initial population have some individuals not respecting bounds OR those said individuals are bounded before executing the evolution.
**For sbx crossover impl**
1. A bounding step at the end of the crossover, making sure neither C1 or C2 are NaN/infinity values and then assigned to genes

 - OS: Ubuntu 20.04
 - Installation method: source
 - Version: branch master

**Additional context**
I have not verified if other genetic algorithms have this problem nor if other crossover operators can assign NaN/infinity values. But this should be confirmed into the same pull request in my opinion.

jschueller:
see #567

### DIFF
--- a/src/algorithms/nsga2.cpp
+++ b/src/algorithms/nsga2.cpp
@@ -128,6 +128,19 @@ population nsga2::evolve(population pop) const
     }
     // ---------------------------------------------------------------------------------------------------------
 
+    // We check that the initial population individuals are within the problem bounds.
+    for (decltype(NP) i = 0u; i < NP; ++i) {
+        const auto &x = pop.get_x()[i];
+        for (decltype(x.size()) j = 0u; j < x.size(); ++j) {
+            if (!std::isfinite(x[j]) || x[j] < bounds.first[j] || x[j] > bounds.second[j]) {
+                pagmo_throw(std::invalid_argument,
+                            "Individual " + std::to_string(i) + " has a gene at position "
+                                + std::to_string(j) + " that is outside the problem bounds. "
+                                + get_name() + " cannot deal with it.");
+            }
+        }
+    }
+
     // No throws, all valid: we clear the logs
     m_log.clear();
 
--- a/src/utils/genetic_operators.cpp
+++ b/src/utils/genetic_operators.cpp
@@ -112,10 +112,8 @@ std::pair<vector_double, vector_double> sbx_crossover_impl(const vector_double &
                 betaq = sbx_betaq(beta, eta_c, rand01);
                 c2 = 0.5 * ((y1 + y2) + betaq * (y2 - y1));
 
-                if (c1 < lb[i]) c1 = lb[i];
-                if (c2 < lb[i]) c2 = lb[i];
-                if (c1 > ub[i]) c1 = ub[i];
-                if (c2 > ub[i]) c2 = ub[i];
+                c1 = std::isfinite(c1) ? std::clamp(c1, lb[i], ub[i]) : lb[i];
+                c2 = std::isfinite(c2) ? std::clamp(c2, lb[i], ub[i]) : lb[i];
                 if (drng(random_engine) < .5) {
                     child1[i] = c1;
                     child2[i] = c2;
--- a/tests/genetic_operators.cpp
+++ b/tests/genetic_operators.cpp
@@ -120,6 +120,54 @@ BOOST_AUTO_TEST_CASE(sbx_crossover_test)
         });
 }
 
+BOOST_AUTO_TEST_CASE(sbx_crossover_nan_inf_test)
+{
+    detail::random_engine_type random_engine(32u);
+    // Parents with genes severely violating bounds can cause a negative beta in the
+    // SBX crossover calculation, which leads to NaN in std::pow(). We verify the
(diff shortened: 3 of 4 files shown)


sylvesterkaczmarek (commented):
The new initial-population bounds check still accepts `NaN`, because both `NaN < lower` and `NaN > upper` are false. Reject non-finite genes explicitly before the range comparison and add a NaN/Inf population regression so invalid decision variables cannot enter NSGA-II.

jschueller:
@sylvesterkaczmarek done

sylvesterkaczmarek (approved):
The initial population check now rejects non-finite genes explicitly before the range comparison, and the tests cover NaN and infinities in addition to ordinary out-of-bounds values. This resolves my review finding.

sylvesterkaczmarek (approved):
Rechecked current a388ab7e. The initial-population validation still rejects non-finite genes before range comparison, and the regression covers NaN plus both infinities alongside ordinary lower/upper-bound violations. The current CI matrix is green. No remaining blocker from my review.

darioizzo:
@jschueller 

1) is the decision of assigning the lower bound to *infs* or *nans* justified ... why not random?
2) whats the fundamental reason nsga2 cannot work with an initial population outside problem bounds?

jschueller:
1. lower bound choice is an arbitrary fail-safe to avoid consuming the rng for determinism
2. its an assumption in the formulas: beta=1+2(y1-yl)/(y2-y1) then pow becomes nan and  fitness breaks sorting

darioizzo:
Ok but if SBX is fixed and will always return finite values, why is NSGA-II requesting a strict lb, ub population? Or maybe I am missing something?

jschueller:
The SBX guard only covers c1/c2 when crossover fires -- child=parent copy keeps the OOB gene when crossover is skipped, the integer part is just swapped, and polynomial_mutation delta=(y-yl)/(yu-yl) also goes NaN via pow(negative) if y is OOB. So the fix only stops nan propagation, it doesn't repair: OOB would still be evaluated and collapsed to lb, biasing search. Fail-fast in evolve is the correct contract since SBX/mutation assume feasible parents.

darioizzo:
I can see the SBX guard establishing certainty on that util to return finite values when pow or anything else fails. Not the NSGAII introducing a request on the initial population to be within bounds in the preamble. Dring the evolve loop anyhthing can be produced and evaluated anyway and the check would be escaped, to only return if the user calls successive evolve.

sylvesterkaczmarek:
One concrete point from tracing the current `evolve()` path: the initial bounds check establishes an invariant that the loop itself preserves. With feasible parents, skipped SBX genes are unchanged feasible parent values, crossed continuous genes are clamped to `[lb, ub]`, integer crossover only swaps feasible parent integers, polynomial mutation starts from `delta1/delta2 in [0,1]` and clamps the result, and integer mutation samples directly from the declared bounds. So I do not see an in-loop path from a valid starting population to an out-of-bounds child in the current implementation. The new non-finite fallback in SBX still makes sense defensively for direct utility calls or externally invalid parents, but the preamble check is what prevents those invalid parents entering NSGA-II in the first place.

darioizzo:
Then I would keep and document the nsgaii new contract and not you touch the sbx impl which would silently change behaviour.

sylvesterkaczmarek:
Agreed. Given the invariant above, keeping the new NSGA-II input contract explicit while leaving the public SBX utility behavior unchanged is cleaner. The preamble validation is sufficient to protect NSGA-II, and dropping the non-finite fallback avoids an unrelated silent semantic change in direct SBX calls.

sylvesterkaczmarek (dismissed):
Rechecked current `a388ab7e` after the maintainer discussion. The head still contains the `sbx_crossover_impl` semantic change from `c520510b` that maps non-finite crossover results to the lower bound, plus its direct SBX regression test. That no longer matches the agreed direction. Please drop the SBX behavior change and its corresponding direct-utility test, keep the NSGA-II initial-population validation/regressions, and document the new NSGA-II contract that initial genes must be finite and within problem bounds. CI is green, but the current diff does not yet reflect the agreed contract.

darioizzo:
@sylvesterkaczmarek  Thanks for your inputs. To keep things clear, please hold off on steering or providing reviews unless explicitly asked. The discussion with @jschueller is still ongoing and no decision has been made yet.

@jschueller  Could you share your view on including the evolve preamble but not the defensive SBX change? Unless you need it for specific reasons, we could defer the SBX proposed change until there’s a stronger use case.

jschueller:
done, keeping only the bounds check + doc

Dave-Poissant:
Opened pull request #568 could be closed then?

PR #631, opened 2026-05-21
refactor: reduce scope of some variables

vil02:


darioizzo:
Explanation missing. Closing. happy to reconsider if well explained/motivated.

PR #636, opened 2026-08-31
Fix installed Ipopt consumer support

BillyONeal:
This fixes three issues with consuming an installed pagmo build configured with `PAGMO_WITH_IPOPT`:

- The installed package config now requests both the Ipopt headers and library, matching the build-time dependency.
- The imported Ipopt library target propagates LAPACK and platform dynamic-loader libraries needed by static consumers.
- `ipopt.hpp` accepts the unprefixed, `coin-or/`, and `coin/` header layouts supported by the finder, so consumers that include only the installation's root include directory can compile it.

This was found while validating microsoft/vcpkg#53599. I tested dynamic and static Ipopt-enabled installs on Windows with Release and Debug CMake consumers, plus direct consumers using only the triplet root include directory and all root libraries.


BillyONeal:
Sorry the bots are on the loose

PR #634, opened 2026-08-23
Fix BFE test for C++20 stateless lambdas

shin4141:
Fixes #632

## Summary

C++20 changed the default-constructibility of stateless lambda closure
types. As a result, a stateless lambda that does not satisfy pagmo's
UDBFE requirements under C++17 can satisfy them under C++20.

The existing BFE test assumes unconditionally that the lambda is not a
UDBFE, while `is_udbfe` itself correctly classifies the callable from
its actual type properties.

This change:

- makes the lambda test follow the callable's default-constructibility
  under the selected C++ standard;
- verifies the other UDBFE requirements independently;
- when the lambda is admitted, verifies that it can be stored, copied
  and invoked through `pagmo::bfe`;
- updates the BFE documentation so it no longer states that lambdas are
  categorically non-default-constructible.

No production BFE or type-trait implementation is changed.

## Verification

Using Apple clang 21.0.0 on arm64 macOS:

- C++17 Debug focused BFE test: PASS
- C++17 Debug enabled core suite: 81/81 PASS
- C++20 Debug focused BFE test: PASS
- C++20 Release enabled core suite: 81/81 PASS

A negative-control run restoring the old unconditional
`!is_udbfe<decltype(fun)>` expectation fails under C++20 as expected.

The C++20 Debug full build is blocked independently by an existing
`ATOMIC_VAR_INIT` deprecation promoted to an error by AppleClang 21;
the focused BFE test passes in that configuration.

I did not reproduce the issue's exact GCC 15.3 / GCC 16.1 Linux
environment locally.

PR #638, opened 2026-09-17
tests: fix bfe test failure under C++20 (Fixes #632)

### LINKED ISSUE #632: [BUG] bfe tests do not succeed under c++ 20
**Describe the bug**
The `bfe` test does not succeed when using the C++ 20 language standard. In particular, when running tests under C++20, running tests fails with
```
 8/81 Test  #8: bfe ..............................***Failed    0.02 sec
Running 11 test cases...
BFE name: Default batch fitness evaluator
        C++ class name: pagmo::default_bfe

        Thread safety: basic

/home/whispers/scratchpad/pagmo2/tests/bfe.cpp(463): error: in "lambda_std_function": check !is_udbfe<decltype(fun)>::value has failed

*** 1 failure is detected in the test module "bfe_test"
```
This is particularly problematic for GCC 16, which changes the default standard version to C++20.

All other tests pass. This same issue is "fixed" by building the project with C++17, and all the tests pass. I'm not very knowledgeable on C++, so I'm not completely sure why this is. The failing test is this:

https://github.com/esa/pagmo2/blob/4aead6ca38d3c0157166d6f1f0070c4312574f94/tests/bfe.cpp#L458-L468

Looking through the code at

https://github.com/esa/pagmo2/blob/4aead6ca38d3c0157166d6f1f0070c4312574f94/include/pagmo/bfe.hpp#L96-L108

and a list of changes for C++20, I think it *might* be because C++20 switches whether stateless lambdas are default constructible. In particular,

```cpp
#include <ios>
#include <iostream>

int main() {
    auto fun = [](int x) { return x + 1; };
    std::cout << std::boolalpha << std::is_default_constructible_v<decltype(fun)> << '\n';
    return 0;
}
```
prints `false` with `g++ -std=c++17 test.cpp && ./a.out` and `true` with `g++ -std=c++20 test.cpp && ./a.out`. I might be completely off base with this, as I don't know C++ very well, but that's my best guess. I'm not certain whether this is a bug in the code or the test; I'm not familiar enough with pagmo to determine. 

**To Reproduce**
1. Clone the repository at https://github.com/esa/pagmo2.
2. Configure the project with `cmake -B build -DPAGMO_BUILD_TESTS=ON -DCMAKE_CXX_STANDARD=20`.
3. Build the project with `make -C build -j16`. 
4. Run tests for this project with `CTEST_OUTPUT_ON_FAILURE=1 make -C build test`.

**Expected behavior**
I expected to see all tests pass, including the "bfe" test.

**Screenshots**
n/a

 - OS: Linux 7.1.4
 - Installation method: Cloning from GitHub and compiling
 - Version: latest master, i.e. commit 4aead6ca38d3c0157166d6f1f0070c4312574f94
 - Compiler versions: GCC 15.3.0 and GCC 16.1.0

**Additional context**
n/a

philphauler:
Hey 👋

Fixes #632. In C++20, stateless lambdas are default-constructible ([P0624R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0624r2.pdf)), which makes a lambda with the correct call signature satisfy all `is_udbfe` requirements. The existing test expects the lambda NOT to be a valid UDBFE, but that was only true in C++17.

The fix guards the assertion with a `__cplusplus` version check:
- **C++20+:** `is_udbfe<decltype(fun)>` is `true` (lambda is default-constructible, so it IS a valid UDBFE)
- **C++17:** `is_udbfe<decltype(fun)>` is `false` (unchanged behavior)

Also adds a C++20 test that constructs a `bfe` from the bare lambda and exercises it, confirming the type trait result is correct for the standard in use.

This unblocks GCC 16, which defaults to C++20.

The MSVC check (`_MSVC_LANG`) is included because MSVC does not always set `__cplusplus` correctly without `/Zc:__cplusplus`.

Tested conceptually against the C++20 spec; no local Boost installation to compile-verify. CI should confirm 🙏

All the best ⚡


darioizzo:
Thanks for submitting this fix and for confirming the C++20 behavior.

We decided to proceed with PR #634 because it addresses the issue at the appropriate scope: the is_udbfe implementation is behaving correctly, while the failing assertion was based on C++17 lambda behavior. PR #634 updates the test to account for the standard-dependent behavior and also corrects the related documentation.

PR #638 reaches the same test result, but PR #634 provides the more complete fix by covering both the test and documentation while avoiding changes to production code.

philphauler:
Sounds totally fine!
I’ll work on more soon


Best regards,
Phil Phauler

*Stay up to date with my latest events and music*
Instagram: https://www.instagram.com/philphauler
Facebook: https://www.facebook.com/PhilPhauler
Youtube: http://bit.ly/philphaulerytb
Soundcloud: https://www.soundcloud.com/phil-phauler-official
Twitter: https://www.twitter.com/philphauler

*Bookings worldwide*
***@***.***
www.wacegroup.com <https://wacegroup.com/artist/phil-phauler/>

*Disclaimer*
This communication contains information which is confidential and may also
be privileged. It is for the exclusive use of the intended recipient(s). If
you are not the intended recipient(s) please note that any distribution,
copying or use of this communication or the information in it is strictly
prohibited. If you have received this communication in error, please notify
us and delete the email and any copies of it.Thank you for your cooperation.

